PR #85, opened 2018-04-16
Added user feedback when saving settings

DalderupMaurice:
Added user feedback when changing settings

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -2,12 +2,15 @@ import React from 'react';
 import { func, string } from 'prop-types';
 
 import styles from './Settings.scss';
+import Saved from '../Saved/Saved';
 
 export default class Settings extends React.Component {
   static propTypes = {
     currentNetwork: string.isRequired,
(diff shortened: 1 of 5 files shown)


src/components/Settings/Settings.js
```diff
@@ -28,11 +32,16 @@ export default class Settings extends React.Component {
             <option value="nOSLocal">nOS Local</option>
           </select>
         </label>
+        {this.state.saved && <Saved /> }
```
mhuggins: Just thinking out loud here.  I think we could do this without the need for setting state.  When the current network changes, the store has progress change from `LOADING` to `LOADED`.  I think we could watch for that change using `withProgress` combined with `componentDidUpdate` checking for a change in progress and updating the view accordingly.

With that said, it's not a fleshed out idea, and I wouldn't let it block merging this.

DalderupMaurice: After what I'm doing now I'll have a look into it 😄 Haven't used withProgress yet

PR #89, opened 2018-04-17
Added the possibility to login using keystore json file

GwendalBrossard:
This PR adds the possibility for the user to login using the keystore json file generated when he created an account via the client.

A new tab (Json File) is now present on the login page:
new interface

It uses react-file-reader (https://www.npmjs.com/package/react-file-reader) for handling html file input.


src/components/Login/Login.js
```diff
@@ -12,11 +13,13 @@ import styles from './Login.scss';
...
 const TAB_PASSPHRASE = 'passphrase';
 const TAB_LEDGER = 'ledger';
+const TAB_JSONFILE = 'jsonfile';
 
 const TABS = {
   [TAB_PASSPHRASE]: 'Passphrase',
   [TAB_WIF]: 'WIF',
-  [TAB_LEDGER]: 'Ledger'
+  [TAB_LEDGER]: 'Ledger',
+  [TAB_JSONFILE]: 'Json File'
```
mhuggins: I think we should call this "Wallet File" for clarity.

src/components/Login/Login.js
```diff
@@ -57,6 +60,8 @@ export default class Login extends React.Component {
         return <LoginFormPassphrase disabled={loading} onLogin={login} />;
       case TAB_LEDGER:
         return <LoginFormLedger disabled={loading} onLogin={login} />;
+      case TAB_JSONFILE:
+        return <LoginFormJsonFile disabled={loading} onLogin={login} />;
```
mhuggins: And name these `TAB_WALLET_FILE`, `LoginFormWalletFile` to match.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
```diff
@@ -0,0 +1,65 @@
+
+import React from 'react';
+import { bool, string, func } from 'prop-types'; import { noop } from 'lodash';
+
+import ReactFileReader from 'react-file-reader'
```
mhuggins: I'd prefer we not add extra dependencies that aren't needed, especially considering that we have access to electron and fs, which makes this quite simple.  You can see an example of how we're saving newly created accounts to a wallet file via the [SaveAccount component](https://github.com/nos/client/blob/master/src/components/Register/AccountDetails/SaveAccount/SaveAccount.js#L55-L75).  We should be able to inverse this logic pretty easily, I think.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
```diff
@@ -0,0 +1,65 @@
...
+
+  static defaultProps = {
+    disabled: false,
+    wif: '',
+    setWIF: noop,
+    onLogin: noop
+  };
+
+
+handleFiles = files => {
```
mhuggins: :art: Looks like the whitespace and alignment is off on some of these lines.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
```diff
@@ -0,0 +1,65 @@
...
+    const { setWIF } = this.props;
+
+    var reader = new FileReader();
+    reader.onload = e =>
+    {
+    let keyStoreJSON = JSON.parse(reader.result);
```
mhuggins: We can actually just let neon-js load the account data from file.  See here: https://github.com/CityOfZion/neon-js/blob/a0967658535ac70361b749e9ba56742c3b2a5766/src/wallet/Wallet.js#L96

There's also an [`accountToNEP6`](https://github.com/nos/client/blob/242b71ca9f70eea23c40e7c799ab41e2f4914c91/src/util/accountToNEP6.js) file in this codebase, it would be nice to include an `accountFromNEP6` counterpart.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
```diff
@@ -0,0 +1,65 @@
...
+
+    let accounts = keyStoreJSON.accounts.map(accounts =>
+      {
+        return accounts.key
+      });
+
+    setWIF(accounts[0]);
```
mhuggins: A wallet can have more than one account in here.  I see you're just pulling the first account.  I think a better experience would be to let the user pick which account they wanted (the accounts have labels that we can display), and then login with that.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
```diff
@@ -0,0 +1,65 @@
...
+  render() {
+    const { disabled } = this.props;
+
+    return (
+      <div className={styles.actions}>
+        <p>Upload your Keystore Json file.</p>
+        <ReactFileReader  fileTypes={'.json'}
+                          multipleFiles={false}
+                          handleFiles={this.handleFiles}>
+            <Button type="submit" disabled={disabled}>Upload</Button>
```
mhuggins: Maybe "Load Wallet" or something instead since the user's not uploading?

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
mhuggins: We also need to handle the error condition here for an invalid file structure.

mhuggins:
Also, I'll mention that the team would be happy to help address any feedback I left here.  Just let us know how you prefer to approach. 😄

src/components/Login/Login.js
GwendalBrossard: OK, changed locally

src/components/Login/Login.js
GwendalBrossard: OK, changed locally

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
GwendalBrossard: OK, changed locally.
I am now using electron to load the wallet file and fs to read the data in it instead.

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
GwendalBrossard: OK, changed locally.
Good point indeed!

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
GwendalBrossard: I can do that but I need some clarification because, currently, the file generated when the  user creates an account via the client and saves the wallet file to its computer stays untouched afterwards.
In this case, there can only be 1 account in the wallet file. This is why I only check for the first value of the array as it should be the only value available.

However, maybe we can add the possibility for the user to save the account in an already existing wallet file. In this case, I can iterate through all the accounts in the wallet file and add the possibility for the user to select an account from all the accounts available in the wallet file.

What do you think of that?

mhuggins: It's true that we're storing only a single account per wallet file at the moment.  However, the NEP6 standard permits multiple accounts per wallet file.  If someone tried to open a NEP6 file from Neon Wallet for example, we shouldn't just assume they exported one account from that app.

mhuggins: We can also just load the first account for now, and an enhancement would be to let the user choose.  I'm okay with separating that work.  We should handle errors though (no accounts or JSON format is invalid).

src/components/Login/LoginFormJsonFile/LoginFormJsonFile.js
deanpress: "Select Wallet File" would make sense imo

deanpress:
Agree with all suggestions by @mhuggins. And we should indeed let users select their account after uploading the JSON file. Let's fix those things and merge this baby!

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.scss
```diff
@@ -4,4 +4,12 @@
...
   }
+
+  .listWallet{
+    width: 100%;
+
+    padding-top: 16px;
+    // float:left;
+  }
+
 }
```
GwendalBrossard: As seen with @deanpress this will change once a class for the <select> will be available.

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -0,0 +1,136 @@
...
+
+    if (!this.walletFile.accounts.length > 0)
+    {
+      this.props.alert(`This wallet file has no account in it.`);
+      return
+    }
+
+    this.removeEverythingFromArray(accountsInWalletFile);
+
+    this.walletFile.accounts.forEach(function(account, i) {
```
DalderupMaurice: I'd use something like this,  `this.setState({ accountsInWalletFile: Object.assign({}, this.walletFile.accounts) })` copying the whole array to the state instead of looping over it and mutating the state

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -0,0 +1,136 @@
...
+
+
+  handleSubmit = (event: Object) => {
+    const { onLogin } = this.props;
+    const { wif } = this.state
+    event.preventDefault()
+    onLogin({ wif });
+  }
+
+  removeEverythingFromArray(array)
```
DalderupMaurice: You can use lodash's function `_.remove` to clear the array. This prevents mutating the state. (It's included already in the dependencies)

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -0,0 +1,136 @@
...
+      if (selectedFile) {
+        this.load(selectedFile[0]); //There can only be 1 value in the array as only 1 file can be loaded at a time
+      }
+    });
+  }
+
+
+  load = (selectedFile) => {
+    const { accountsInWalletFile } = this.state;
+    // const { walletFile } = this.props
```
DalderupMaurice: This is not needed anymore?

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -0,0 +1,136 @@
...
+      <form className={styles.loginForm} onSubmit={this.handleSubmit}>
+        <Button onClick={this.handleLoadWallet} disabled={disabled}>Select Wallet File</Button>
+
+        <select
+          className={styles.listWallet}
+          value={wif}
+          onChange={this.handleAccountSelected}
+        >
+        <option value=''>Select an account</option>
+          {accountsInWalletFile.map((account, index) => {
```
DalderupMaurice: index not used? :P

DalderupMaurice:
It's a great PR overall 😄 Happy to see the contribution!

GwendalBrossard:
Thank you for your review @DalderupMaurice 
I have made the changes locally and I am waiting for the `Select` class before pushing the changes :)

src/components/Login/LoginFormWalletFile/LoginFormWalletFile.js
mhuggins: Prefer ES6 syntax for that sample code though:

```js
this.setState({ accountsInWalletFile: { ...this.walletFile.accounts } })
```

jeroenptrs:
@deanpress @mhuggins @GwendalBrossard @DalderupMaurice added Select component from our style guide. Feel free to add changes / PR! 🙌

DalderupMaurice:
Seems good to me, wbu? @mhuggins

mhuggins:
Thanks for the contribution and for working with our feedback! 😄 👏

Issue #82, opened 2018-04-14
Change `getBalance` to take an address


mhuggins:
We currently have `getBalance` in our API, which takes an asset ID or NEP5 script hash as an argument and retrieves the current account's balance for that asset/token.

I suggest we change this function to take two arguments, with the first argument being the address that we should fetch the balance for.  This would make it more extensive in case another address balance is needed, and it could still work smoothly with the way our API works, e.g.:

```js
nos.getAddress()
  .then((address) => nos.getBalance(address, NEO))
  .then((balance) => console.log(balance))
  .catch((err) => console.log(err.message));
```

jeroenptrs:
That's pretty good, let's open up a branch

PR #91, opened 2018-04-18
feat(balances): extended the balances API 

DalderupMaurice:
Extended the balances API with an optional parameter to query other people their balance
Solves issue #82

### DIFF
--- a/src/actions/dapps/makeBalancesAction.js
+++ b/src/actions/dapps/makeBalancesAction.js
@@ -1,0 +1,10 @@
+import { createActions } from 'spunky';
+
+import getBalances from '../../util/getBalances';
+
+export const ID = 'balances_api';
+
+export default createActions(ID, ({ net, address, tokens = [], addedAddress } = {}) =>
+  async () => {
+    return getBalances({ net, address, tokens, addedAddress });
+  });
--- a/src/actions/dapps/makeBalancesAction.js
+++ b/src/actions/dapps/makeBalancesAction.js
@@ -1,10 +1,14 @@
 import { createActions } from 'spunky';
 
 import getBalances from '../../util/getBalances';
+import generateDAppActionId from './generateDAppActionId';
 
-export const ID = 'balances_api';
+export const ID = 'balances';
 
-export default createActions(ID, ({ net, address, tokens = [], addedAddress } = {}) =>
-  async () => {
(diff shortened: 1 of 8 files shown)


src/actions/dapps/makeBalancesAction.js
```diff
@@ -0,0 +1,10 @@
+import { createActions } from 'spunky';
+
+import getBalances from '../../util/getBalances';
+
+export const ID = 'balances_api';
+
+export default createActions(ID, ({ net, address, tokens = [], addedAddress } = {}) =>
```
mhuggins: The structure of this doesn't match the other action-makers under `actions/dapps`.  This should be wrapped in a function that takes a `sessionId` and `requestId`, which then builds a unique identifier for the redux store. i.e.:

```js
export const ID = 'balances';

export default function makeBalancesActions(sessionId, requestId) {
  const id = generateDAppActionId(sessionId, `${ID}-${requestId}`);

  return createActions(id, ({ net, address, tokens = [] }) => () => {
    return getBalances({ net, address, tokens })
  });
}
```

src/components/Browser/RequestsProcessor/GetBalance/index.js
```diff
@@ -15,10 +15,18 @@ export default function makeGetBalance() {
   return compose(
```
mhuggins: You'll need to update `makeGetBalances` to accept `balancesActions` instead, and then within `mappings.js` include `makeBalancesActions` there.

DalderupMaurice:
Added PR comments although using the 2nd parameters (giving an address to fetch it's balance) does return 0. @mhuggins was going to have a look at the issue

src/actions/dapps/makeBalancesAction.js
```diff
@@ -9,6 +10,10 @@ export default function makeBalancesActions(sessionId, requestId) {
   const id = generateDAppActionId(sessionId, `${ID}-${requestId}`);
 
   return createActions(id, ({ net, address, tokens = [] } = {}) => async () => {
+    if (!wallet.isAddress(address)) {
+      throw new Error(`Invalid address: "${address}".`);
+    }
```
DalderupMaurice: This is already done in getBalances 😄

mhuggins: Oops!  I'll drop remove this commit.

PR #92, opened 2018-04-19
Added `nos` protocol & use it to resolve queries

mhuggins:
This replaces the `nameServiceActions` with a way of resolving the `nos` protocol.  This allows us to highjack not just queries and requests entered into the address bar, but other sources in the page (e.g.: images, scripts, links, anchors, etc.).


public/resolve.js
```diff
@@ -0,0 +1,42 @@
+const { rpc, u } = require('@cityofzion/neon-js');
+
+// TODO: Configurable network settings and script hash
```
DalderupMaurice: PR Looks good to me besides this TODO, but it can be merged (after the CI Build is fixed) 
We can add this in later

mhuggins: Fixed the build, merging!

PR #96, opened 2018-04-22
Fix/building issues

iidarko:
Try to clean up and fix some issues


jeroenptrs:
This branch isn't ready yet. As you can see our tests are failing and we need to figure out why. In the mean time I'm going to close this pull.

PR #105, opened 2018-04-25
Add Node.js LTS version in nvmrc file

robmoorman:


DalderupMaurice:
I'd prefer defining the engine versions in the package.json instead of adding yet another file to the root folder

robmoorman:
Ok, so we would like to provide a bandwith to node engine versions (8.11.x series all work). I'll do some testing with multiple version and provide an engine settings here.

jeroenptrs:
I'm adding the node version to a branch, will merge this in that branch.
@robmoorman might want to set it to >= 8.11.1 and < 9.0.0, to support future LTS updates

robmoorman:
We may want to ignore this pr @jeroenptrs in this case, as nvm only uses strict node version. It does support 'lts/*' versions, but this doesn't work well...

jeroenptrs:
@robmoorman alright. I will close this PR then and will get to work on defining node versions in the package. I do want to stress that you should use the PR template next time by the way 😉

PR #106, opened 2018-04-25
No-drag fix

XaroRSA:
There was a problem when loading the registration page in the application, where a user is unable to select/click the input fields.
This PR fixes that issue.


DalderupMaurice:
What code is actually implemented to fix the error, could you please point this out?
As I look at the code, it's a bunch of old commits thrown in 1 PR.
Could you also fill in the PR template? It's here for a reason.

XaroRSA:
Sorry something not right here. I will start over.

DalderupMaurice:
No problem 😄 
Glad to see people putting in an effort

Issue #86, opened 2018-04-17
Add additional neoscan URLs in Network Settings


deanpress:
Currently users can only select networks from a set of predetermined neoscan URLs. We should let users add custom ones.

robmoorman:
Entering an invalid url is your own responsibility I would say. Keeping up your (custom) list in Redux would be the way to go. Maybe prefill (as initial state) this with the ones currently hardcoded.

kitchenSpoon:
Thanks @robmoorman , would asking for a name and a neoscan address, like in this page be enough?
http://cityofzion.io/neon-js/docs/en/basic_privnet.html

kitchenSpoon:
Is there anything else that needs to be done to move this from In progress to Done?

PR #109, opened 2018-04-26
feat(settings): Add additional custom url network configurations

kitchenSpoon:
## Description

I've added the ability to add and save your own custom network configurations( custom url )

## Motivation and Context

Currently users can only select networks from a set of predetermined neoscan URLs. We should let users add custom ones.
https://github.com/nos/client/issues/86

screen shot 2018-04-26 at 11 25 35 pm

## Types of changes

## Checklist:

## Closing issues

Fixes #


DalderupMaurice:
The CI is failing, could you have a look at it? Besides that, nice PR thank you!
(I am not sure if you can see the details of the CI? Let me know!)

jeroenptrs:
@kitchenSpoon thanks for your contribution! it's greatly appreciated 💪 
Might want to take a look at CiCi's tests, might give you an indication why the tests are failing https://circleci.com/gh/nos/client/349?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Don't forget to mock network requests where appropriate!

kitchenSpoon:
@DalderupMaurice @jeroenptrs Thanks guys, I've managed to fix the Failing test, please let me know if theres anything else I need to change. I come from a native iOS/Android background so I'm not too familiar with much of the elements and css so do let me know of any improvements.

src/hocs/withCustomNetworkDialog.js
```diff
@@ -0,0 +1,11 @@
...
+
+import { confirm } from '../actions/dialogsActions';
+
+export default function withCustomNetworkDialog(propName = 'confirm') {
+  const mapDispatchToProps = (dispatch) => ({
+    [propName]: (...args) => dispatch(confirm(...args))
+  });
+
+  return connect(null, mapDispatchToProps);
+}
```
DalderupMaurice: Is this used? I don't see any references to it in the code

kitchenSpoon: Yep, its not used, I'll removed it

src/actions/settings/currentNetworkActions.js
```diff
@@ -2,13 +2,45 @@ import { createActions } from 'spunky';
 import { getStorage, setStorage } from '../../lib/storage';
 
 export const ID = 'currentNetwork';
+export const NETWORKS_ID = 'networks';
```
mhuggins: This constant and all related functions should go into a new file called `networksActions` since it's completely unrelated data that it's storing.

src/actions/settings/currentNetworkActions.js
```diff
@@ -2,13 +2,45 @@ import { createActions } from 'spunky';
...
 export default createActions(ID, () => async () => {
   const currentNetwork = await getStorage(ID);
-  return typeof currentNetwork === 'string' ? currentNetwork : 'TestNet';
+  if (typeof currentNetwork !== 'string') {
+    return 'TestNet';
+  }
+  return currentNetwork;
+});
+
+export const getAllNetworks = createActions(NETWORKS_ID, () => async () => {
```
mhuggins: I would consider making this the default export in the new file this gets moved to, but I don't feel strongly either way.

src/components/Settings/Settings.js
```diff
@@ -1,13 +1,25 @@
 import React from 'react';
-import { func, string } from 'prop-types';
+import { func, string, object, arrayOf } from 'prop-types';
 
+import { noop } from 'redux-saga/utils';
```
mhuggins: :neckbeard: Super minor (non-blocking), and not something our linting is currently configured to catch -- in all other files, our imports are organized as:

1. group of by absolute package imports
2. whitespace
3. grouped by relative app imports

In that vein, this line should go above the whitespace rather than below.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
         {this.state.saved && <Saved /> }
       </div>
     );
   }
 
+  renderButtons = () => {
+    return (
+      <div>
+        <Button onClick={this.handleAddNewNetwork}>
+          {'Add custom network configuration'}
```
mhuggins: :neckbeard: No need to interpolate the string here.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+        </Button>
+        <div className={styles.divider} />
+        <Button onClick={this.handleClearNetwork}>
+          {'Clear custom network configurations'}
```
mhuggins: :neckbeard: No need to interpolate the string here.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+                  >
+                    {network.name}
+                  </option>
+                );
+              })
+            }
           </select>
         </label>
+        <div className={styles.buttonContainer}>
+          Custom Neoscan Url: {this.props.currentNetwork}
```
mhuggins: 🔤 Capitalize "URL" here.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
             <option value="TestNet">TestNet</option>
             <option value="CozNet">CozNet</option>
             <option value="nOSLocal">nOS Local</option>
+
+            {
+              this.props.allNetworks.map((network) => {
+                return (
+                  <option
+                    key={network.neoscan}
```
mhuggins: Should the key be `network.name` instead?  It seems like it could be possible for two networks to point to the same neoscan URL, which would make these non-unique.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+                    value={network.neoscan}
+                  >
+                    {network.name}
+                  </option>
+                );
+              })
+            }
           </select>
         </label>
+        <div className={styles.buttonContainer}>
```
mhuggins: `buttonContainer` doesn't seem like the right name here since it's not containing any buttons, rather it just has text.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+          {'Clear custom network configurations'}
+        </Button>
+      </div>
+    );
+  }
+
+  handleClearNetwork = () => {
+    this.props.clearNetworks();
+    this.props.setCurrentNetwork('TestNet');
+    this.props.alert('All custom network configurations cleared');
```
mhuggins: 🔤 All our other alerts are proper sentences that end with a period, I think we should be consistent for user-facing messages.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+        <Button onClick={this.handleClearNetwork}>
+          {'Clear custom network configurations'}
+        </Button>
+      </div>
+    );
+  }
+
+  handleClearNetwork = () => {
+    this.props.clearNetworks();
+    this.props.setCurrentNetwork('TestNet');
```
mhuggins: Do we always want to switch to TestNet when the networks are cleared?  It seems like we should only do that if the current network is one that's being cleared.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+          label="Network URL"
+          placeholder="Network URL"
+          onChange={this.handleChangeNetworkUrl}
+        />
+      </div>
+    ), {
+      title: 'New network configuration',
+      onConfirm: this.handleConfirmAddNetwork,
+      onCancel: noop
+    });
```
mhuggins: 🤔 Not something we have to address for this PR, but food for thought -- I'd like to come up with a way that we can encapsulate confirmation dialogs like this to prevent the parent component from rendering complex form details and managing modal state.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+  }
+
+  handleConfirmAddNetwork = () => {
+    const network = this.props.allNetworks.find((element) => {
+      return element.neoscan === this.props.networkUrl;
```
mhuggins: Similar concern here as with the `key` comment from earlier.  Should we search by name instead?  The name seems like it should be unique while the neoscan URL doesn't necessarily have to be.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+    });
+
+    if (network) {
+      this.props.alert('Error: A network configuration with that Url already exist');
```
mhuggins: 🔤 Capitalize "URL" and end sentence with a period.  (The capitalization may be a moot point based upon my prior comment.)

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+      return;
+    }
+
+    const newNetwork = { name: this.props.networkName, neoscan: this.props.networkUrl };
+    this.props.addNetwork(newNetwork);
+    this.props.setCurrentNetwork(this.props.networkUrl);
```
mhuggins: We've been storing the network as "MainNet", "TestNet", etc. previously.  It looks like this will pass in a URL instead now.  I think we should pass in the name instead.  I think this aligns with all my other name-related comments.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+      case 'CozNet':
+      case 'nOSLocal':
+        return this.saveNetwork(event.target.value);
+      default:
+        break;
+    }
+
+    // Custom networks
+    const network = this.props.allNetworks.find((element) => {
+      return element.neoscan === event.target.value;
```
mhuggins: Same comment here as with others.

src/components/Settings/Settings.js
```diff
@@ -30,18 +42,122 @@ export default class Settings extends React.Component {
...
+    });
+    return this.saveNetwork(network.neoscan);
```
mhuggins: `network.name` instead of `network.neoscan`.

src/hocs/withNetworkData.js
```diff
@@ -1,8 +1,15 @@
...
+  const mapSettingsDataToProps = (currentNetwork) => ({
+    [propName]: currentNetwork
+  });
   return withData(currentNetworkActions, mapSettingsDataToProps);
 }
+
+export function withAllNetworkData() {
+  const mapSettingsDataToProps = (allNetworks) => ({ allNetworks });
+  return withData(getAllNetworks, mapSettingsDataToProps);
+}
```
mhuggins: This should go into its own `withAllNetworksData.js` HOC file.

mhuggins:
Please let me/the team know if you need any assistance with addressing the feedback, we'll be happy to help!

src/components/Settings/Settings.js
DalderupMaurice: This is exactly what I need in the other open PR, we should have this as a portal (or HOC?) for easy reusability in another PR :P

DalderupMaurice:
Awesome work man! If these are addressed we're ready to merge ;) 
If we need to help out or help with some tasks let us know!

kitchenSpoon:
Thanks for the comments guys! I'll check them out and see if I can them all sorted out and get to back you all.

src/components/Settings/Settings.js
kitchenSpoon: Yup I agree, I feel bad having the parent component managing state for the modal, but struggle to come up with some way to have the modal manage its own state and be generic so that its reusable in the way like the confirm and alerts modal are. What be happen to discuss more.

src/components/Settings/Settings.js
kitchenSpoon: Previously I have experimented with this
```
 export default function withNetworkData(propName = 'net') {
   const mapSettingsDataToProps = (currentNetwork) => ({
-    name: currentNetwork.name, [propName]: currentNetwork.neoscan
+    [propName]: currentNetwork.neoscan
   });
   return withData(currentNetworkActions, mapSettingsDataToProps);
 }

```

src/components/Settings/Settings.js
mhuggins: I'll add an issue to track this.  Thanks for the feedback, guys.

mhuggins: Added #118.

src/components/Settings/Settings.js
mhuggins: We can either allow the user to provide neoscan URL as part of their config, or we can just point neoscan to "http://localhost:4000/api/main_net" maybe as a starting point.  I don't think we have to dynamically pass in the network name to neoscan since that won't work.

kitchenSpoon: For the first option, does that mean that the user cannot add new neoscan URL after their config? (at the beginning) and for the second option, it feels like its actually not completing the PR, and in fact im not sure how we can point neoscan to "http://localhost:4000/api/main_net"  without saving the url of "http://localhost:4000/api/main_net" .

kitchenSpoon: @mhuggins if I have to save the network name instead of the URL, then i might have to do the mapping from name to URL somewhere in currentNetworkActions.js, would that be okay? I'll code up an implementation after work.

src/components/Settings/Settings.js
```diff
@@ -26,22 +40,119 @@ export default class Settings extends React.Component {
...
           >
-            <option value="MainNet">MainNet</option>
-            <option value="TestNet">TestNet</option>
-            <option value="CozNet">CozNet</option>
-            <option value="nOSLocal">nOS Local</option>
+            {map(settings.networks, this.renderNetworkOption)}
           </select>
         </label>
+        <div className={styles.networkDetails}>
+          Custom Neoscan URL: {this.getCurrentNetworkUrl()}
```
kitchenSpoon: @mhuggins I've added this back because now we are storing network name and the bottom part of the app is showing the name instead of url.

kitchenSpoon: I've added this back because now we are storing network name and the bottom part of the app is showing the name instead of url.

src/components/Settings/Settings.js
```diff
@@ -26,22 +40,119 @@ export default class Settings extends React.Component {
...
+          Clear custom network configurations
+        </Button>
+      </div>
+    );
+  }
+
+  handleClearNetworks = () => {
+    // If user is on a custom network, switch to TestNet
+    if (!PREDEFINED_NETWORKS.includes(this.props.currentNetwork)) {
+      this.props.setCurrentNetwork('TestNet');
```
kitchenSpoon: I've also changed this to `setCurrentNetwork` instead of `saveNetwork` because we are not saving and the `saveNetwork` code will trigger setState while the alert for clearing network is shown which causes an error `Can't call setState (or forceUpdate) on an unmounted component`

PR #111, opened 2018-04-26
feat(register): save the new account to an already existing Wallet file

GwendalBrossard:
## Description

This PR adds the possibility for the users, after creating a new account via the client, to choose between saving this account as a new NP6 wallet (the only option available at the moment) or to add it to an already existing wallet (the feature this PR adds).

## Motivation and Context

Currently, the users, after creating a new account via the registration page, can only choose to export the account by creating a new NP6. I think it would be nice to be able to add the account to a wallet of their choices.
Before: Only one button 'Save' was available on the account details page.
Now: Two buttons are available 'Save as new NEP6 Wallet' on the account details page.

## How Has This Been Tested?

Tested locally with:
- A wallet with 1+ accounts -> OK
- A wallet with 1 account -> OK
- A wallet with no account -> OK
- An empty file -> Error message pops-up (OK)

feat register

## Types of changes

## Checklist:

## Closing issues

Fixes #

### DIFF
--- a/src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
+++ b/src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
@@ -100,7 +100,7 @@ export default class SaveAccount extends React.Component {
       return;
     }
 
-    const newAccount = new wallet.Account({ ...account, label });
+    const newAccount = new wallet.Account({ ...account, label, isDefault: true });
     const newWallet = new wallet.Wallet({ accounts: [newAccount] });
     await this.save(filename, newWallet);
   }
--- a/src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
+++ b/src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
@@ -9,7 +9,6 @@ import { wallet } from '@cityofzion/neon-js';
 import Input from '../../../Forms/Input';
 import Button from '../../../Forms/Button';
 import accountShape from '../../../../shapes/accountShape';
-import accountToNEP6 from '../../../../util/accountToNEP6';
 import styles from './SaveAccount.scss';
 
 const writeFile = promisify(fs.writeFile);
@@ -40,7 +39,6 @@ export default class SaveAccount extends React.Component {
         />
         <div className={styles.saveButtons}>
           <Button
-            id={styles.buttonNew}
             className={styles.button}
             disabled={isEmpty(this.props.label)}
             onClick={this.handleSaveNewWallet}
@@ -48,7 +46,6 @@ export default class SaveAccount extends React.Component {
             Save as new NP6 Wallet
           </Button>
           <Button
-            id={styles.buttonAdd}
             className={styles.button}
             disabled={isEmpty(this.props.label)}
             onClick={this.handleAddToWallet}
@@ -65,7 +62,7 @@ export default class SaveAccount extends React.Component {
   }
 
 
-  handleAddToWallet = () => {
+  handleAddToWallet = async () => {
     const { account, label } = this.props;
 
     const filenames = remote.dialog.showOpenDialog({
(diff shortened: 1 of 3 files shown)


src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
```diff
@@ -37,13 +38,24 @@ export default class SaveAccount extends React.Component {
...
-        <Button
-          className={styles.button}
-          disabled={isEmpty(this.props.label)}
-          onClick={this.handleSave}
-        >
-          Save
-        </Button>
+        <div className={styles.saveButtons}>
+          <Button
+            id={styles.buttonNew}
```
mhuggins: We shouldn't use `id` for styling, just `className`.  I think we can style these buttons the way you want without providing distinct names, I'll add some comments to the `scss` file to clarify.

src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
```diff
@@ -52,9 +64,32 @@ export default class SaveAccount extends React.Component {
...
+  handleAddToWallet = () => {
+    const { account, label } = this.props;
+
+    const filenames = remote.dialog.showOpenDialog({
+      title: 'Add account to a NEP6 Wallet',
+      message: 'Add account to a NEP6 Wallet',
+      filters: [{ name: 'NEP6 Wallet File', extensions: ['json'] }]
+    });
+
+    if (filenames) {
```
mhuggins: :neckbeard: To prevent nested `if`'s, it might be a little cleaner to return earlier here.  i.e.:

```js
if (!filenames) {
  return;
}
```

src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
```diff
@@ -52,9 +64,32 @@ export default class SaveAccount extends React.Component {
...
+      const walletLoaded = this.loadWallet(filenames[0]);
+
+      if (walletLoaded != null) {
+        const newAccount = new wallet.Account({ ...account, label });
+        walletLoaded.addAccount(newAccount);
+        this.save(filenames[0], walletLoaded);
```
mhuggins: Missing an `await` here.

src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
```diff
@@ -63,12 +98,37 @@ export default class SaveAccount extends React.Component {
...
+    }
+  }
+
+
+  save = async (filename, walletLoaded = null) => {
     const { label, account } = this.props;
-    const data = JSON.stringify(accountToNEP6({ ...account, label }));
+    let data = null;
+
+    if (walletLoaded == null) {
```
mhuggins: This conditional seems odd for this function.  I would suggest changing the existing `this.save(...)` call from `handleSaveNewWallet` to instead pass in `walletLoaded` with its representative wallet instance.

src/components/Register/AccountDetails/SaveAccount/SaveAccount.scss
```diff
@@ -1,14 +1,19 @@
...
-    margin: 0 0 16px 16px;
+    width: 49%;
+    margin: 0 0 16px 0px;
+  }
+
+  #buttonNew {
+    float: left;
+  }
+  #buttonAdd {
+    float: right;
```
mhuggins: We should stick with flexbox instead of using floats.  ([This](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) is a great resource if you've never used it.)  I haven't tested this, but I think this would work:

```js
.saveButtons {
  display: flex;
  align-items: center;
  align-content: stretch;
  flex-wrap: nowrap;

  .button {
    flex: 1 1 auto;

    &:first-child {
      margin-right: 8px;
    }

    &:last-child {
      margin-right: 8px;
    }
  }
}
```

GwendalBrossard: I'm not familiar with flexbox, I'm going to read that, thanks.

PR #117, opened 2018-04-28
Added nos protocol & use it to resolve queries

mhuggins:
## Description
This replaces the `nameServiceActions` with a way of resolving the `nos` protocol.  This allows us to highjack not just queries and requests entered into the address bar, but other sources in the page (e.g.: images, scripts, links, anchors, etc.).

## Motivation and Context
* By adding a custom protocol, we can use the built in `URL` class to parse queries for us.
* By adding a protocol handler, we can support dApps referencing images, sources, and other third-party files stored via the `nos` protocol.  (e.g.: ``)
* By adding a protocol handler, we're a step closer to native OS support.  (e.g.: Users can click an `nos://` link from within Chrome.)
* By moving this logic further away from the redux action, we get closer to supporting tabbed browsing.

## How Has This Been Tested?
Smoke testing.

NOTE: There's still an issue with dapp-starter-kit not working, which is why this is marked as "work-in-progress".

TODO.

## Types of changes

## Checklist:

## Closing issues
N/A


jeroenptrs:
@mhuggins re webview crash: does it crash silently? Or do we have some errors that can help us out?

mhuggins:
@jeroenptrs It's crashing, but I haven't been able to track down the cause of the crash.  Been scouring electron docs to see how to debug, but haven't found anything that's worked for me yet.

mhuggins:
It only crashed when I try to load `nos://localhost:1234` while the dapp-starter-kit is running.  If I comment out the `<script>` tag in the starter kit, then it loads fine.  So there's something going on with trying to load up JS or just that JS in particular.

mhuggins:
Alright, so the issues I had here actually came down to using Parcel with HMR enabled when using a custom protocol.  I replicated this in a [new repo](https://github.com/nos/electron-parcel-custom-protocol) and [opened an issue with electron](https://github.com/electron/electron/issues/12855).

As a workaround, we're able to have dapp-starter-kit use the `--no-hmr` CLI option, which we weren't making use of anyway.  I've opened https://github.com/nos/create-nos-dapp/pull/8 to address this.

PR #122, opened 2018-05-02
chore(accessibility): add focus-ring management

indrekpaas:
## Description
Improved keyboard accessibility with native focus-rings and dedicated focus-ring management.

## Motivation and Context
Native focus-rings are currently disabled in favor of visual aesthetics. Unfortunately, this creates a serious accessibility concern since keyboard users rely on focus-rings to track their current position within a document.

The updated behavior shows native focus-rings for keyboard navigation and hides them for mouse input. This is achieved with a popular micro-library [what-input](https://github.com/ten1seven/what-input).

## How Has This Been Tested?
Tested manually and with `yarn test`.

## Types of changes

## Checklist:


mhuggins:
Thanks for the PR!  I agree that `outline: none` is counter to the idea that things should be accessible.  I'm not familiar with the package you're including here, so I'd like to pull this branch down and test it locally before approving.  In general, this is probably a good change though.

mhuggins:
I pulled this down and tried it.  The idea of improving accessibility is great.  However, it looks like this only gets us partway there.  As an example, on the login screen, I used the tab key to focus on the "WIF" and "Leger" tabs, but pressing <kbd>Enter</kbd> and <kbd>Space</kbd> did nothing.  Is there a way to customize what-input to make that happen globally, or do we need to add `tabIndex` and `role` properties to everything still?

mhuggins:
@indrekpaas Please bear with me.  I've never used what-input, and from everything I've read in the package, I'm not completely understanding how this gets us closer to accessibility.

Is the benefit of this package just to ensure that we only show focus rights when the keyboard is used, but not for other device interactions?  If that's the case, then let me know, and I'm cool with accepting/merging this.

indrekpaas:
@mhuggins Sorry if I unintentionally came across as being arrogant.

What-input was created out of a need, where many developers disabled focus-rings because of their (often unaesthetic) appearance, without considering accessibility implications. What-input offers the best of both worlds in that sense.

This feature is complete and future-proof. Only thing to note from hereon out is to keep CSS’ `outline` properties intact, unless providing an alternative styling for focus-rings (like the buttons and text-inputs currently do).

Happy to help with other accessibility improvements across the project. Cheers!

mhuggins:
> Sorry if I unintentionally came across as being arrogant.

Oh, you weren't!  I just wanted to make sure I fully understood.  This looks great, thanks!

Issue #27, opened 2018-03-31
Create Account and Saved Accounts (Encrypted)


deanpress:
On the authentication window, the Passphrase tab should provide the option for users to create new accounts, and to save them on creation or log-in.

Similar to `neon-wallet`, users should be able to select previously saved accounts (only encrypted keys) so they only have to enter their Passphrase to log in.

mhuggins:
This is done!

mhuggins:
My mistake -- we're saving the wallet file, but we're not providing login with a prepopulated dropdown of accounts anywhere.  I believe that's what the intention of this issue is, is that right @deanpress?

kitchenSpoon:
@deanpress , @mhuggins we do have the functionality on the Wallet file Tab though

<img width="1362" alt="screen shot 2018-05-05 at 1 49 27 am" src="https://user-images.githubusercontent.com/2775497/39637495-92738fb6-5006-11e8-9bd5-8e526b459d1b.png">
![Uploading Screen Shot 2018-05-05 at 1.49.23 am.png…]()

mhuggins:
This is unencrypted though.  It's not ideal to have a plaintext wallet file sitting on your computer.  We may still want to support storing encrypted keys that depend on passphrase, which can then prepopulate in a `select` within the "passphrase" tab.

kitchenSpoon:
Ahh, I see thanks

DalderupMaurice:
What's the status of this issue? @mhuggins 
KitchenSpoon's PR is merged already :P

mhuggins:
Closing!

PR #127, opened 2018-05-06
feat(accounts): Login with saved encrypted account

kitchenSpoon:
## Description

Instead of saving the private key in the wallet file, save the encrypted key instead. Also asked for a passphrase when login using wallet file now that we are storing the encrypted key instead of private key. This is also backwards compatibility with wallet file that saves account with unencrypted private key.

This does not solve all the issues in but should be a step the right direction.
https://github.com/nos/client/issues/27

## Motivation and Context

It is not safe to store private key in a file. Saving the encrypted key instead is safer.

## How Has This Been Tested?

Tested locally.

screen shot 2018-05-12 at 12 16 02 am
screen shot 2018-05-12 at 12 15 58 am

## Types of changes

## Checklist:

## Closing issues

Fixes #


src/components/Register/AccountDetails/SaveAccount/SaveAccount.js
```diff
@@ -84,7 +84,7 @@ export default class SaveAccount extends React.Component {
       return;
     }
 
-    const newAccount = new wallet.Account({ ...account, label });
+    const newAccount = new wallet.Account({ ...account, key: account.encryptedKey, label });
```
mhuggins: I don't think we can change this and still have it considered a NEP6 wallet.  The [proposal](https://github.com/neo-project/proposals/blob/master/nep-6.mediawiki#account) states:

* `key` is the private key of the account in the NEP-2 format. This field can be null (for watch-only address or non-standard address).

An alternative to this would be to store encrypted keys to local storage once they are entered in the "Passphrase" tab, and allow those to appear in a dropdown in that tab after being used once.

kitchenSpoon: Hmm ok I get what you are saying, I'll go with your alternative method. Although i feel like there may be some errors with neon.js though in Account.js, it seems like the export function is taking the `_encrypted` variable and exporting it as the variable `key`, should it pehaps export it as `encryptedKey`? Is it because the AccountLike object does not have a `encryptedKey` field?
```
/**
   * Export Account as a AccountLike object.
   * @return {AccountLike}
   */
  export () {
    let key = null
    if (this._privateKey && !this._encrypted) throw new Error('Encrypt private key first!')
    if (this._encrypted) key = this._encrypted
    return {
      address: this.address,
      label: this.label,
      isDefault: this.isDefault,
      lock: this.lock,
      key,
      contract: this.contract,
      extra: this.extra
    }
  }
```

mhuggins: Thanks for calling that out.  Perhaps I am unclear on the what the NEP6 proposal is saying.  I pinged snowypowers on Discord to ask about this since he wrote this code.  I'm interested in what he has to say.

If it's true that we can just use an encrypted version here, then this seems okay to me.  But it sounds like we would need to account for each account in the wallet being either a private key or an encrypted key.

mhuggins: @kitchenSpoon Alright, I got some insight from snowy.

> NEP6 is always encrypted key or null (for watch only addresses)

So it seems we should have behavior for saving the wallet that's more in line with your original change here.  (I haven't gone through the additional commits yet.)

mhuggins: @kitchenSpoon It looks like we can use `wallet.isNEP2` to see if it's encrypted.  Found [here](http://cityofzion.io/neon-js/docs/en/api-wallet.html#verify).

kitchenSpoon: @mhuggins thanks!, let me know how it goes with snowypowers when he replies you.

kitchenSpoon:
@mhuggins I've added back my original changes and also checks for whether its a private key or a encrypted key. I've added a description to let user know whether the account selected is using a private key or encrypted key so they are not confused with why sometimes the passphrase input field appears.

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
       return null;
     }
 
     return (
-      <Select className={styles.accounts} value={wif} onChange={this.handleSelect}>
-        <option value="">Select an account</option>
-        {map(this.props.accounts, (account, index) => (
-          <option value={account.encrypted} key={`account${index}`}>{account.label}</option>
-        ))}
-      </Select>
```
mhuggins: I think we should have this function just return the `<Select>` as it was before, and call `this.renderPassphraseInput()` and `this.renderDescription()` directly from `render`.

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
+      <div>
+        <Select className={styles.accounts} value={encryptedWIF} onChange={this.handleSelect}>
```
mhuggins: This admittedly was missing before you worked on it, but would you mind adding `disabled={disabled}` (based upon `this.props.disabled`) in case the login action is processing?

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
+    let input;
+    if (encryptedWIF !== '' && !wallet.isPrivateKey(encryptedWIF)) {
+      input = (
+        <Input
+          id="passphrase"
+          type="password"
+          label="Passphrase"
+          placeholder="Enter passphrase"
+          value={passphrase}
+          onChange={this.handleChangePassphrase}
```
mhuggins: We need `disabled={disabled}` on this component as well.

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
+        {this.renderDescription()}
+      </div>
     );
   }
 
+  renderPassphraseInput = () => {
+    const { encryptedWIF, passphrase } = this.props;
+
+    let input;
+    if (encryptedWIF !== '' && !wallet.isPrivateKey(encryptedWIF)) {
```
mhuggins: 💅 Can we drop the `let` and inverse the logic here to simply return early if needed?  i.e.:

```js
if (encryptedWIF === '' || wallet.isPrivateKey(encryptedWIF)) {
  return null;
}

return (
  <Input ... />
);
```

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
+      );
+    }
+    return input;
+  }
+
+  renderDescription = () => {
+    const { encryptedWIF } = this.props;
+
+    if (encryptedWIF === '') {
+      return '';
```
mhuggins: Prefer `return null` here.

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -47,22 +52,59 @@ export default class LoginFormWalletFile extends React.Component {
...
+    }
+
+    if (wallet.isPrivateKey(encryptedWIF)) {
+      return 'Private key detected';
+    } else {
+      return 'Encrypted key detected, please type passphrase';
```
mhuggins: Can we add punctuation to the end of both of these?

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -76,19 +118,31 @@ export default class LoginFormWalletFile extends React.Component {
 
   handleSubmit = (event) => {
     event.preventDefault();
-    this.props.onLogin({ wif: this.props.wif });
+    const { encryptedWIF, passphrase = '', onLogin } = this.props;
```
mhuggins: You shouldn't need to assign `passphrase = ''`.  If it's not a string here, there's a bug.

src/login/components/LoginFormWalletFile/LoginFormWalletFile.js
```diff
@@ -76,19 +118,31 @@ export default class LoginFormWalletFile extends React.Component {
...
+      encryptedWIF,
+      passphrase
+    };
+
+    onLogin(loginCredentials);
   }
 
   handleSelect = (event) => {
-    this.props.setWIF(event.target.value);
+    this.props.setEncryptedWIF(event.target.value);
```
mhuggins: Let's also reset the passphrase when a new item is selected from the dropdown.

```js
this.props.setPassphrase('')
```

jeroenptrs:
Great job @kitchenSpoon !

PR #132, opened 2018-05-09
chore(api): return full result object instead of script

jeroenptrs:
## Description
testInvoke now returns the full result object instead of just the script.

## Motivation and Context
* Give developers more info on their invocations (gas consumed)
* Allow developers that want to leverage the actual calculated result to do so

## How Has This Been Tested?
* Fired up the starter-kit, testInvoke now returns the result object. Validated that childs like `gas_consumed` and `stack` were present.

screenshot 2018-05-09 09 50 03

## Types of changes

## Checklist:

## Closing issues

Fixes #

### DIFF
--- a/src/browser/components/RequestsProcessor/TestInvoke/TestInvoke.js
+++ b/src/browser/components/RequestsProcessor/TestInvoke/TestInvoke.js
@@ -1,9 +1,11 @@
 import React from 'react';
-import { any, objectOf, func } from 'prop-types';
+import { func } from 'prop-types';
+
+import testInvokeShape from '../../../shapes/testInvokeShape';
 
 export default class TestInvoke extends React.Component {
   static propTypes = {
-    result: objectOf(any).isRequired,
+    result: testInvokeShape.isRequired,
     onResolve: func.isRequired
   };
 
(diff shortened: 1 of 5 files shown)


src/browser/components/RequestsProcessor/TestInvoke/TestInvoke.js
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
-import { string, func } from 'prop-types';
+import { any, objectOf, func } from 'prop-types';
 
 export default class TestInvoke extends React.Component {
   static propTypes = {
-    script: string.isRequired,
+    result: objectOf(any).isRequired,
```
DalderupMaurice: Should we make a shape for this? @mhuggins

jeroenptrs: Possibly. As per [this tutorial](http://cityofzion.io/neon-js/docs/en/basic_createscript.html#simple-calls-with-createscript) it would resemble something like:
```
shape({
  script: string.isRequired,
  state: string.isRequired,
  gas_consumed: string.isRequired,
  stack: arrayOf(any).isRequired
});
```

and we can even further breakdown the stack array, since it has this shape:
```
shape({
  type: string.isRequired,
  value: string.isRequired
});
```

I'll put those in.

jeroenptrs: Shapes added @DalderupMaurice

PR #140, opened 2018-05-11
feat(splash): Updated splash screen

### LINKED ISSUE #130: Improve the Loading Splash Screen with animated text and UI tweaks
## Current Behavior
The current Loading splash screen does not indicate responsiveness and feels clunky due to the size and window controls.

## Possible Solution
1. Change the Loading splash screen size to 275px x 330px
2. Remove the Window controls
3. Display animated text (with entertaining phrases) that changes with random 2-4 second intervals to indicate responsiveness.
4. Below the animated text, display individual loading steps.

Reference of good loading splash (Discord):
Loading GIF

Soudny:
## Description
Made changes to the splash screen that appears when first loading the application.

This PR has changed the size of splash screen container, removed the header and changed the animation to be a SVG animation. Currently, the animation is a simple loading circle that changes colors and is recommended to be updated with some design team's help.

## Motivation and Context
It has been requested to have splashscreen contain a bit more complex animations that are more eye-pleasing. For that reason, CSS seems like the hard-way to go. Gifs and videos are not so good because they tend to weight a lot - which is what brings us to SVG animations. A simple json file that any designer can create in Adobe After Effects can be imported by simply swapping out the data.json file in the public folder.

The changes were made according to the following issue: https://github.com/nos/client/issues/130

## How Has This Been Tested?
/

image

## Types of changes

## Checklist:

## Closing issues
Fixes #130


public/splash.html
```diff
@@ -4,51 +4,80 @@
...
 
+    <script>
+      if (typeof module === 'object') {
+        // unset module
+        window.module = module;
+        module = undefined;
+      }
+    </script>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/bodymovin/4.13.0/bodymovin_light.min.js"></script>
```
mhuggins: I'm not familiar with this package and will take a look at it, but if we want to use a third party library here, I suggest we:

1. Include the [npm package](https://www.npmjs.com/package/bodymovin) in our package.json instead.
2. Change this script tag to use the npm package.

public/splash.html
```diff
@@ -4,51 +4,80 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 
+    <script>
+      if (typeof module === 'object') {
+        // unset module
+        window.module = module;
+        module = undefined;
```
mhuggins: Can you explain this in more detail and why it's needed?

public/splash.html
```diff
@@ -4,51 +4,80 @@
...
+      }
 
     </style>
 
   </head>
   <body>
 
-    <div class="bg cn"><div class="loader"></div></div>
+    <div class="bg cn">
+        <div id="bm"></div>
```
mhuggins: Can we call this `animation` or something more useful instead of `bm`?

public/animations/loader.js
```diff
@@ -0,0 +1,12 @@
...
+ * The command is tied to script-imported library in DOM,
+ * and therefore cannot have a bodymovin variable defined
+ */
+// eslint-disable-next-line
+bodymovin.loadAnimation({
+  container: document.getElementById('bm'),
+  renderer: 'svg',
+  loop: true,
+  autoplay: true,
+  path: 'animations/data.json'
```
mhuggins: Is this a series of multiple svg's that are stitched together to create an animation?

public/splash.html
jeroenptrs: Bodymovin is pretty great, but I'm not sure it's necessary.

mhuggins: @jeroenptrs My assumption (without having tried yet) is that bodymovin will let us have an animation that's more than just something spinning in a circle.  So it might depend on what the actual animation we're going to use is.  (Unless you have a specific alternative to bodymovin that makes sense, I'm definitely interested in other feedback here.)

public/splash.html
Soudny: Electron does not give its window module executable options that we are used to seeing in browser. If there wasn’t for this wrapper around function and variable definitions, bodymovin would be undefined. Same would happen if you were to add jquery or any other dependency

public/splash.html
Soudny: Yes, bodymovin is actually a powerful tool. You might have heard of lottie for mobile - well this is basically it for web. Even the npm package is called lottie-web. 
In the future, this will allow very, very complex animations to run in a very lightweight and resolution-independant way. Also, the design teams should love this addition as it is “the thing” right now.

I agree with the idea of using npm package, however if we were to do it that way, we’d have to change rendering to go through some templating engine such as ejs or handlebars - simply to inject something from node_modules.

Maybe you know of an easier way?

Btw lottie-web should eventually definitely go into package.json as I see ot commonly used

public/splash.html
Soudny: Sure, makes sense

public/splash.html
mhuggins: I'm working on another branch that should allow us to use npm packages from these HTML files by compiling through webpack.  With that said, it's not ready yet, and so I'm not worried about making it work quite yet.

mhuggins: I just tried this branch locally, and it's a nice improvement.  I don't think we should merge it with a dependency on a remote file though.  Two ideas are:

1. install the npm package, then copy the JS from `node_modules` into `build` as part of the webpack compilation,
2. download the JS file and directly save it to the `public` folder.

For both of these suggestions, we would have to update the `script` tag to reference the local file instead.

@jeroenptrs What do you think about this?  I know you had a separate opinion about not using bodymovin, which may be more valid.

jeroenptrs: @mhuggins Not really, my concern was about adding another dependency that will only be used for one animation. Really like your idea!

Soudny: Sounds good! I'll update the script file to reference it in the `public` folder and, just to make things a bit simpler, add `public/animations` as eslint exceptions (2/3 files are exceptions, so don't see why not)

mhuggins:
@Soudny One more question for you.  If we want to replace the animation at some point, does this depend on AfterEffects to do so?

Soudny:
Not really, AfterEffects is simply the most commonly used method of getting svg animations, but there are plenty of both free and paid options to work with in order to produce the same results. As I already said, the options are endless and we can look well beyond simple loaders

Soudny:
I have added bodymovin minimized dependency to public folder and changed the ID of the animation element. Feel free to take a look and try it out again when you get the chance

DalderupMaurice:
LGTM, wbu? @mhuggins

PR #159, opened 2018-05-20
feat(app): nOS API refactor

jeroenptrs:
## Description
* nOS API functions that require multiple arguments are now called with a `config` object that holds all the arguments. These include: `invoke`, `testInvoke`, `getStorage`, `send`, `getBalance`
* added `encodeArgs` for `testInvoke` and `invoke`, `doInvoke` already does this out of the box apparently. (defaults truthy)
* refactored the `getStorage` `encode` flag to `encodeInput` (defaults truthy)
* added `decodeOutput` flag to `getStorage` (defaults truthy)

## Motivation and Context
Highly needed for encoding checks and future additional configurations.

## How Has This Been Tested?
Tested this for each function, used neosense for SC specific functions.

## Types of changes
Refactoring brought up breaking changes, encodeArgs is a feature.

## Checklist:

- [x] Docs have been added / updated (for bug fixes / features)


Mickieki:
H

README.md
```diff
@@ -46,7 +46,7 @@ We welcome contributions to the code base. If you are interested in becoming a c
 - [Code contribution guidelines](/.github/CONTRIBUTING.md#Code-Contribution)
 
 
-There is a [specific channel called develop](https://discord.nos.io/) on Discord to discuss development.
+There is a [specific channel called develop](https://discord.gg/CXZb3BS) on Discord to discuss development.
```
mhuggins: How come we're changing this link?  The existing one seems to work fine and matches the Discord linked from https://nos.io.

docs/api.md
```diff
@@ -99,100 +98,104 @@ nos.claimGas()
...
 authenticated account.  It does not require the user to grant permission.
 
-### Parameters
-1. `string` - The script hash of the Smart Contract you want to invoke.
-2. `string` - The operation of the Smart Contract you want to invoke.
-3. `...any` - Zero or more args of the Smart Contract you want to invoke.
+#### Parameters
+1. `scriptHash` [string] - The script hash of the Smart Contract you want to invoke.
+2. `operation` [string] - The operation of the Smart Contract you want to invoke.
+3. `args` [string] - An arguments array of the Smart Contract you want to invoke.
```
mhuggins: I think we should change `[string]` to `[string[]]` or `[array<string>]`.

docs/api.md
```diff
@@ -50,47 +50,46 @@ nos.getAddress()
...
+### `getBalance`
 The `getBalance` function provides the balance of a certain address for a
 specified asset or NEP5 token.  It does not require the user to grant permission.
 
-### Parameters
-1. `string` - The asset ID or NEP5 token script hash.
-2. `string` (Optional) - The address of the user you'd like to receive the balance for. This defaults to the currently logged on user if the parameter is not passed.
+#### Parameters
+1. `asset` [string] - The asset ID or NEP5 token script hash.
+2. `address` [string] (Optional) - The address of the user you'd like to receive the balance for. This defaults to the currently logged on user if the parameter is not passed.
```
mhuggins: I like the way snowy writes the parameters out here for options/config objects, and I'm interested to see how you feel about it.  It'd end up looking something like this:

```
#### Parameters
* `config` [object] - The config options to perform this operation.
* `config.asset` [string] - The asset ID or NEP5 token script hash.
* `config.address` [string] (Optional) - The address of the user you'd like to receive the balance for. This defaults to the currently logged on user if the parameter is not passed.
```

docs/api.md
```diff
@@ -99,100 +98,104 @@ nos.claimGas()
...
 
-### Parameters
-1. `string` - The asset ID script hash.
-1. `string` - The amount of the asset to send.  It is recommended that strings are used instead of
-   numbers to prevent floating point rounding issues.
-1. `string` - The recipient address of the asset.
+#### Parameters
+1. `asset` [string] - The asset ID script hash.
+2. `amount` [string] - The amount of the asset to send.
+**NOTE: It is recommended that strings are used instead of numbers to prevent floating point rounding issues.**
```
mhuggins: I think we need to either append this to the previous line or indent it several spaces to prevent the ordered list from being interrupted.

src/browser/util/encodeArgs.js
```diff
@@ -0,0 +1,30 @@
...
+  // For now this takes care of strings and numbers and nested Arrays
+  // Also this shouldn't be used when NEO addresses are passed for contracts using CheckWitness.
+  // Until we determine what a good practice for general encoding addresses is
+  // - using unhexlify + reverseHex + toScriptHash or just reverseHex as is used in this function -
+  // but the former is necessary for contracts using CheckWitness
+  // the latter is used from neon-js guides to checkBalance.
+  // If the former can be used for those as well, this is no problem.
+  const encodedArgs = args.map((arg) => {
+    const executor = tree`
+      lazy ${arg}
```
mhuggins: Is this basically like a fancy switch statement?

src/browser/util/encodeArgs.js
```diff
@@ -0,0 +1,30 @@
...
+        expression ${wallet.isAddress} ${() => u.reverseHex(wallet.getScriptHashFromAddress(arg))}
+        expression ${Array.isArray} ${() => encodeArgs(arg)}
+        type ${'string'} ${() => u.str2hexstring(arg)}
+        type ${1} ${() => u.int2hex(arg)}
```
mhuggins: It's not clear to me here (even after reading the package documentation) why you call `type ${1}` here instead of something like `type ${'number'}`.  Would anything change here if you called `type ${2}` instead?  Maybe I'm interpreting it as the equivalent of JS's `typeof` operator when that's not the case...?

src/browser/util/encodeArgs.js
```diff
@@ -0,0 +1,30 @@
...
+    return args;
+  }
+
+  // For now this takes care of strings and numbers and nested Arrays
+  // Also this shouldn't be used when NEO addresses are passed for contracts using CheckWitness.
+  // Until we determine what a good practice for general encoding addresses is
+  // - using unhexlify + reverseHex + toScriptHash or just reverseHex as is used in this function -
+  // but the former is necessary for contracts using CheckWitness
+  // the latter is used from neon-js guides to checkBalance.
+  // If the former can be used for those as well, this is no problem.
```
mhuggins: :neckbeard: Punctuation is a bit inconsistent throughout this comment, it made it a little difficult to scan since it spans so many lines.  I think adding some periods where needed would help. 😄

src/browser/actions/makeTestInvokeActions.js
```diff
@@ -30,7 +30,7 @@ const testInvoke = async ({ net, scriptHash, operation, args }) => {
 export default function makeTestInvokeActions(sessionId, requestId) {
   const id = generateDAppActionId(sessionId, `${ID}-${requestId}`);
 
-  return createActions(id, ({ net, scriptHash, operation, args }) => () => {
-    return testInvoke({ net, scriptHash, operation, args });
+  return createActions(id, ({ net, scriptHash, operation, args, encodeArgs }) => () => {
```
mhuggins: It looks like `encodeArgs` is being treated as falsy, but is really just `undefined`.  It might be more clearer to specify `encodeArgs = false` here, assuming that's the correct default.

src/browser/actions/makeStorageActions.js
```diff
@@ -16,14 +16,17 @@ const getStorage = async ({ net, scriptHash, key, encode = true }) => {
...
+      .execute(endpoint);
+  return decodeOutput ? u.hexstring2str(result) : result;
 };
 
 export default function makeStorageActions(sessionId, requestId) {
   const id = generateDAppActionId(sessionId, `${ID}-${requestId}`);
 
-  return createActions(id, ({ net, scriptHash, key, encode }) => async () => {
-    return getStorage({ net, scriptHash, key, encode });
+  return createActions(id, ({ net, scriptHash, key, encodeInput, decodeOutput }) => async () => {
```
mhuggins: Similar suggestion here.  Since this is the "public-facing" function (the one that is exported), I think we should specify the defaults here rather than on line 8.

src/browser/actions/makeInvokeActions.js
```diff
@@ -45,7 +45,15 @@ const doInvoke = async ({ net, address, wif, scriptHash, operation, args }) => {
...
-  return createActions(id, ({ net, address, wif, scriptHash, operation, args }) => () => {
-    return doInvoke({ net, address, wif, scriptHash, operation, args });
+  return createActions(id, ({
+    net,
+    address,
+    wif,
+    scriptHash,
+    operation,
+    args,
+    encodeArgs
```
mhuggins: Same here.

src/browser/components/RequestsProcessor/GetBalance/GetBalance.js
```diff
@@ -5,34 +5,34 @@ import { wallet } from '@cityofzion/neon-js';
...
-  return wallet.isScriptHash(scriptHash) || [NEO, GAS].includes(scriptHash);
+function isValidAssetHash(asset) {
+  return wallet.isScriptHash(asset) || [NEO, GAS].includes(asset);
 }
 
 export default class GetBalance extends React.Component {
   static propTypes = {
     balances: arrayOf(balanceShape).isRequired,
-    scriptHash: string.isRequired,
+    asset: string.isRequired,
```
mhuggins: We'll want to support tokens in the future too.  Should we just refer to it as an "asset" always?  I'm okay with it, just want to make sure we're explicitly making that decision.

src/browser/components/RequestsProcessor/GetStorage/index.js
```diff
@@ -17,27 +19,31 @@ export default function makeStorageComponent(storageActions) {
...
     // Rename arguments given by the user
     withProps(({ args }) => {
-      const options = args[2] || {};
-      return {
-        scriptHash: args[0],
-        storageKey: args[1],
-        encode: !!options.encode
-      };
+      const result = pick(args[0], CONFIG_KEYS);
+      result.index = result.key; // key is reserved in React props, so we map it to index
```
mhuggins: Extremely minor nitpick, but if we call `Object.keys(result)` before and after this line, it's possible that "key" won't be in there before, but "index" will be in there after.  The value will still be `undefined` though, so it's probably not worth worrying about here unless we're validating user input in some sense that depends on that behavioral difference.

README.md
jeroenptrs: Yep, however this one is a *direct* invite to develop, which seemed appropriate regarding the context. If you prefer the general discord link, I'll change it back

docs/api.md
jeroenptrs: Yes, good call.

docs/api.md
jeroenptrs: Yeah that's a good idea!

src/browser/util/encodeArgs.js
jeroenptrs: Very fancy 😂 It's a package of my own, yeah 😬 https://github.com/jeroenptrs/switch-tree

src/browser/util/encodeArgs.js
jeroenptrs: The `type` evaluator does typeof x === typeof y. So in this case you're looking for a number, you have to evaluate it against a number. But that's a good suggestion though, I might change it in a later version.

src/browser/components/RequestsProcessor/GetBalance/GetBalance.js
jeroenptrs: This is consistent with `send`, which is why I changed it actually 😄

src/browser/actions/makeStorageActions.js
DalderupMaurice: What's the benefit of this? Can it be seen on autocomplete or such? :P

README.md
mhuggins: This is fine, thanks for explaining

src/browser/actions/makeStorageActions.js
mhuggins: It's just to ensure we're passing the intended values, nothing more.

src/browser/actions/makeTestInvokeActions.js
mhuggins: We're explicitly passing an argument though, which means the default inside of createScript won't ever be used. That's why I suggest moving/adding it here. :smile:

src/browser/components/RequestsProcessor/GetStorage/index.js
mhuggins: More like:

```js
if (result.key) {
  result.index = result.key;
  delete result.key;
}
```

But not really important I don't think.

src/browser/actions/makeTestInvokeActions.js
jeroenptrs: Alright, thanks for explaining! 😄

PR #173, opened 2018-05-26
chore(browser): add responsive tab

### LINKED ISSUE #164: Prevent many browser tabs from growing beyond window width
## Expected Behavior
When "many" browser tabs are opened, they should eventually begin to shrink in the same manner as Chrome.  There should also be some predefined max number of tabs allowed open.

## Current Behavior
When opening many tabs, they cause the window to scroll and reveal our hideous background color.

energizer-tabs

## Possible Solution
Use CSS to specify a set width for tabs, using flexbox to ensure they shrink if needed.

1. Login successfully.
2. Open many browser tabs.

## Context
We want the browser to be usable and presentable.

## Your Environment

jarindr:
## Description

this PR fixed the tab overflow problem, since the width of tabs is not fixed which make tabs to overflow the view

## Motivation and Context

it solve the problem that tabs can be overflow and does not make flex to do its job
https://github.com/nos/client/issues/164
we need `min-width: 0` for flex shrink to be able to shrink beyond text (white-space:nowrap;)

## How Has This Been Tested?

tested with many screen size and works with most of it (responsive)

may-26-2018 22-09-04

## Types of changes

## Checklist:

## Closing issues

Fixes #164


jeroenptrs:
@mhuggins care to take a look? 👼

src/browser/components/Tabs/Tabs.scss
```diff
@@ -3,11 +3,11 @@
   flex-wrap: nowrap;
   padding: 0 32px;
   border-bottom: 1px solid #dee2e6;
-
+  width: calc(100vw - 64px);
```
mmoshi: @jarindr - is 64px the sidebar width? Looks like a magic number. Was there any reason we didn't just go with 100%?

I'm will work a little on https://github.com/nos/client/issues/187 and might adjust this at the same time.

jarindr: you can either fix the width of the body => to all containers and use 100% width here, but i choose to go with this solution that doesn't need to specify width for all parent containers

mmoshi: @jarindr `100vw` is the entire viewport width, the sidebar is 70px so it looks about right in the tabs, but you're 6px off. It only works because the tabs are `box-sizing: border-box`.

I would rather add a `width: calc(100vw - $authenticatedLayout-menu-width)` rule to the main container, and then the children can all rely on 100% width working as expected.

jarindr: @mmoshi you can do that, or you can remove padding and fix the width of tabs and align center it, if you wanna change that there are many ways :) no right or wrong

jarindr: if you really wanna do it right, don't use `vw` but start from the body and the containers inherit the width instead

jarindr: also i think you are right that it was off by 6px didn't check it was `box-sizing: border-box`

PR #177, opened 2018-05-26
feat(browser): remove request queue when logout

### LINKED ISSUE #162: Clear pending requests when a session ends
## Expected Behavior
When a browser session ends (via browser tab closure or signing out), the request queue for that session should be cleared, and no pending actions (e.g.: transactions, address fetches, balance fetches, etc.) should be processed.

## Current Behavior
When a browser session ends, the request queue for that session is not cleared.

## Possible Solution
When changing tabs:
* `requestsReducer` should listen for the `CLOSE_TAB` action, and behave the same as the `EMPTY_REQUESTS` action (for the specified `sessionId`).

When logging out:
* Create a new `emptyAll` action in `requestsActions` and handler in `requestsReducer`.  This action should remove all pending requests regardless of `sessionId`.
* Add `withLogout` HOC to the `Logout` container, and have it dispatch the `emptyAll` action defined above.

1. Run dapp-starter-kit
2. Open nos://localhost:1234 within the nOS client
3. Perform an asynchronous request
4. Very quickly switch tabs (depends on #160 being merged) or sign out.

## Context
If a user signs out, a dapp can still perform pending actions on behalf of the users.  Alternatively, changing accounts means that a dapp could try to perform an action as the next signed in user.

## Your Environment

jarindr:
## Description

add empty queue for every session Id functionality when logout event is fired

## Motivation and Context

when user signs out, a dapp can still perform pending actions on behalf of the users. Alternatively, changing accounts means that a dapp could try to perform an action as the next signed in user

## How Has This Been Tested?

runs dapp and clear session successfully 

## Types of changes

## Checklist:

## Closing issues

Fixes #162


src/browser/actions/requestsActions.js
```diff
@@ -1,7 +1,8 @@
 export const ENQUEUE_REQUEST = 'ENQUEUE_REQUEST';
 export const DEQUEUE_REQUEST = 'DEQUEUE_REQUEST';
 export const EMPTY_REQUESTS = 'EMPTY_REQUESTS';
-
+export const EMPTY_ALL_REQUESTS = 'EMPTY_ALL_REQUESTS';
```
mhuggins: :neckbeard: Can you leave a line of whitespace between the string constants and the function constants?  I think it will improve "scannability" for the file. 😄

src/browser/reducers/requestsReducer.js
```diff
@@ -42,6 +43,10 @@ export default function requestsReducer(state = initialState, action) {
       return dequeue(state, action.sessionId, action.id);
     case EMPTY_REQUESTS:
       return empty(state, action.sessionId);
+    case EMPTY_ALL_REQUESTS:
+      return initialState;
+    case CLOSE_TAB:
```
mhuggins: This can be combined with the `EMPTY_REQUESTS` case, i.e.:

```js
case EMPTY_REQUESTS:
case CLOSE_TAB:
  return empty(state, action.sessionId);
```

src/logout/components/Logout/Logout.js
```diff
@@ -3,10 +3,12 @@ import { func } from 'prop-types';
 
 export default class Logout extends React.Component {
   static propTypes = {
-    logout: func.isRequired
+    logout: func.isRequired,
+    emptyAllRequests: func.isRequired
```
mhuggins: I'd prefer if we continue to only pass a single `logout` function prop in here, and extend what the function does within `logout/components/Logout/index.js`.

src/logout/components/Logout/index.js
```diff
@@ -1,16 +1,20 @@
...
+
 const mapActionsToProps = ({ reset }) => ({
   logout: reset
 });
 
 export default compose(
   withActions(accountActions, mapActionsToProps),
+  connect(null, (dispatch) => ({
+    emptyAllRequests: () => dispatch(emptyAll())
+  })),
```
mhuggins: In conjunction with my previous comment, this `connect` call would need to shift up above `withActions` so that we can reference the `emptyAllRequests` function from within the new `logout` prop.  Something like this:

```js
const mapDispatchToProps = (dispatch) => ({
  emptyAllRequest: () => dispatch(emptyAll())
});

const mapActionsToProps = ({ reset }, { emptyAllRequests }) => ({
  logout: () => {
    reset();
    emptyAllRequests();
  }
});

export default compose(
  connect(null, mapDispatchToProps),
  withAction(accountActions, mapActionsToProps),
  /* ...snip... */
)(Logout);
```

src/browser/reducers/requestsReducer.js
jarindr: if that's how the codebase is 👍

src/logout/components/Logout/index.js
jarindr: 👍

jarindr: I just dig deeper, I don't think what you are trying to do will work?
- withAction works with only one actions from saga and `mapActionsToProps` https://github.com/neoverse/spunky/blob/master/src/hocs/withActions.js
if you really want only one prop which is logout, maybe compose will be a solution

jarindr:
@mhuggins added few comments on your suggest solution

src/logout/components/Logout/index.js
mhuggins: Last piece of feedback.  To be consistent with our other files, can this function be lifted into a `mapDispatchToProps` function before it's used here?

src/logout/components/Logout/index.js
```diff
@@ -1,16 +1,26 @@
...
+
 const mapActionsToProps = ({ reset }) => ({
-  logout: reset
+  reset
 });
 
 export default compose(
   withActions(accountActions, mapActionsToProps),
-  withLogout((state, { history }) => history.push('/login'))
+  connect(null, (dispatch) => ({
```
mhuggins: Last piece of feedback. To be consistent with our other files, can this function be lifted into a `mapDispatchToProps` const before it's used here?

jarindr: should be good now 👍

Issue #150, opened 2018-05-14
NEO app in nano ledger is not required to open to login to nOS


atmahesh8:
NEO app in nano ledger is not required to open to login to nOS, when it is already accessed once. 

## Expected Behavior

nOS ledger tab login button should be disabled and not allowed to login without NEO app in ledger open. 

## Current Behavior

nOS ledger tab login button is enabled and allowing to login without NEO app in ledger open. 
https://drive.google.com/file/d/148Py97VniIeIcczrufzX3yEFHacJ5wMk/view?usp=sharing

## Possible Solution

Disable login button when it is detected that NEO app is not open. 

## Steps to Reproduce (for bugs)

1. Go to Ledger tab in nOS
2.  Connect Ledger Nano S to USB
3.  Open NEO app in Ledger Nano S
4. Click login button in Ledger tab in nOS
5. click logout after login to nOS client
6. close NEO app in Ledger nano S or disconnect Ledger nano S from USB
7. Click login button in Ledger tab in nOS - able to login with previous session(public key is displayed on account tab)

## Context

User should not be allowed to login to multiple apps using NANO ledger S. it's a must-fix.

## Your Environment

* Version used:
* Environment name and version (e.g. node.js 9.6.1, yarn 1.6.1, npm 5.6.0): Yarn 1.5.1 / 1.6.0, Node v8.9.1 / v10.1.0
* Operating System and version (desktop or mobile): macOS Sierra

jeroenptrs:
@atmahesh8 That's weird, but you do have a NEO private key stored on your ledger?

atmahesh8:
Yes, NEO private key are stored in ledger. I could login ETH wallet ledger and nOS client at the same time.

atmahesh8:
In fact, I disconnected by Ledger nano and could login to nOS by just click of login button. I think it is not being logged out properly and retaining the session.

atmahesh8:
@jeroenptrs , can you give some pointers to troubleshoot by looking at logout functionality. I have asked for help on discord as well.

atmahesh8:
found the js files. looking at it. thx

atmahesh8:
Though no Ledger connected ,able to see connected to Ledger. 
<img width="1613" alt="image" src="https://user-images.githubusercontent.com/39273533/40223804-d9a2183a-5aa1-11e8-8bd6-7d286804ffc3.png">

mhuggins:
This has been fixed.

Issue #148, opened 2018-05-14
No USB device found message persists


atmahesh8:
An error message after finding nano ledger - No USB device found.

## Expected Behavior

Only `Found USB Ledger Nano S` should be displayed. 

## Current Behavior

`No USB device found` is also being displayed in addition to `Found USB Ledger Nano S`
<img width="424" alt="screen shot 2018-05-14 at 9 38 50 pm" src="https://user-images.githubusercontent.com/39273533/40011406-61310d4a-57c5-11e8-8f7d-e43ff4d34163.png">

## Possible Solution

Removing `No USB device found` message when detected ledger. 

## Steps to Reproduce (for bugs)

1. Go to Ledger tab in nOS
2. Open NEO in Ledger
3. Ledger tab will display
```Connected to ledger
Found USB Ledger Nano S
No USB device found
```

## Context

No negative side-effects, UI issue

## Your Environment

* Version used:
* Environment name and version (e.g. node.js 9.6.1, yarn 1.6.1, npm 5.6.0): Yarn 1.5.1 / 1.6.0, Node v8.9.1 / v10.1.0
* Operating System and version (desktop or mobile): macOS Sierra
* Ledger Nano S firmware : 1.4.2

jeroenptrs:
@atmahesh8 thank you for all the Ledger related issues you've been opening, it's nice to have someone with a Ledger helping us out with bugs/issues/functionality.

However, please refrain from editing the formatting of the Issue template 😛 makes the issues unreadable. I'll edit your issues and titles to make them a bit more readable.

atmahesh8:
thanks @jeroenptrs . I took a note of formatting.

atmahesh8:
While debugging this issue I found out that it is related to ledger version firmware version. ( update ledger version in the environment) . 

I am now getting three errors randomly when NEO app is open in the ledger. 

1.  `Invalid sequence`
2.  `Cannot open device with path. `( As reported in https://github.com/nos/client/issues/151) 

 <img width="1635" alt="screen shot 2018-05-18 at 12 03 08 am" src="https://user-images.githubusercontent.com/39273533/40196716-fd86e50a-5a2e-11e8-9c54-534df5054a0f.png">
3. `No USB device found`

<img width="1555" alt="screen shot 2018-05-17 at 11 56 27 pm" src="https://user-images.githubusercontent.com/39273533/40196729-04e2ff5a-5a2f-11e8-8fb8-9db97deb0aa0.png">

[...]

PR #186, opened 2018-05-28
chore(login): refactored ledger login by using states

tjuretzek:
## Description
The Ledger login showed inconsistent status texts.
The errors and successfull login persist after interacting and unplugging a hardware ledger.

## Motivation and Context
https://github.com/nos/client/issues/148
https://github.com/nos/client/issues/150

## How Has This Been Tested?
I used the Ledger Nano S and plugged it in to my mac, while the nOS client was open and unplugged it after successful registration at the client to create an error and disable the login.

Status: client started and clicked on 'ledger' login mode (login button disabled)
bildschirmfoto 2018-05-28 um 11 37 50

Status: Plugged in Ledger Nano S and unlocked it (new Message: "Found USB Ledger device", Login button still disabled)
bildschirmfoto 2018-05-28 um 12 01 54

Status: Opened the NEO App on the Ledger Nano S (now the deviceInfo is availlable and the connection is established -> Login button now active)
bildschirmfoto 2018-05-28 um 12 02 06

Status: unplugged the Ledger Nano S (Login is disabled again)
bildschirmfoto 2018-05-28 um 12 02 26

## Types of changes
I redid the functionality of the LoginFormLedger.js. I implemented a logic only calling the status functions if they are needed. I added two new states as a connection of a ledger device indicator and an indicator for the app opened on the ledger.
The status messages were also put into states, so the component updates itself.

## Checklist:

## Closing issues

Fixes #


jeroenptrs:
@tjuretzek Thanks for your contribution!
You might want to update your code to appease our linting rules, however 😄

tjuretzek:
@jeroenptrs 😳ooops - rules are updated and I refactored the code

jeroenptrs:
I don't have a Ledger, so testing will mostly be done by @mhuggins

mhuggins:
@affendy93 Please refrain from adding the same comments across multiple issues, PRs, etc.  It adds a lot of notification noise without contributing to the conversation.  Thanks!

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -42,12 +50,14 @@ export default class LoginFormLedger extends React.Component {
   }
 
   render() {
+    this.getDeviceInfo();
```
mhuggins: Is this call intentional?  Is it trying to change state?  Changing state from within a render call is something that should not be done within React, as it can chain re-renders potentially indefinitely.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -56,50 +66,63 @@ export default class LoginFormLedger extends React.Component {
   }
 
   renderActions = () => {
-    const disabled = this.props.disabled || !this.props.deviceInfo;
+    const disabled = (this.state.connected && this.state.open) ? this.props.disabled : true;
     const onClick = disabled ? null : this.handleLogin;
 
+    this.state.deviceError = (!disabled) ?
```
mhuggins: This is not a valid manner of changing state within React.  We should only ever use `setState`.  (Please see the [React documentation](https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly) for reference.)

With that said, in line with my previous comment, changing React state during rendering should be avoided at all costs.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -56,50 +66,63 @@ export default class LoginFormLedger extends React.Component {
...
-    } else {
-      return <p>Searching for USB devices. Please plug in your Ledger to login.</p>;
-    }
+    event.preventDefault();
+    onLogin({ publicKey });
   }
 
-  renderDeviceStatus = () => {
-    const { publicKey } = this.props;
+  getDeviceInfo = () => {
```
mhuggins: For consistency & clarity, I would prefer that we ensure any function that returns React nodes we should prefix with `render` rather than `get`.

I also think this function (along with `getDeviceStatus` below) is overloaded in what it's doing.  It's attempting to inspect device status, change state, and render the result.  I think it would be easier to follow what's happening if these things remain separate.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -56,50 +66,63 @@ export default class LoginFormLedger extends React.Component {
...
+    const { deviceInfo, deviceError } = this.props;
```
mhuggins: We're using props here, but then moving the props into the state?  I'm not following how this is solving the problem.

tjuretzek:
@mhuggins you were 100% right. I appologize.
I implemented the solution you outlined above and tested it with the nano Ledger S.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -1,34 +1,37 @@
...
+const { FAILED } = progressValues;
+
 const deviceInfoShape = shape({
   manufacturer: string.isRequired,
   product: string.isRequired
 });
 
 export default class LoginFormLedger extends React.Component {
   static propTypes = {
-    disabled: bool,
```
mhuggins: We still want to account for the `disabled` prop.  This is used to disable the login button while login is processing.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -1,34 +1,37 @@
...
   };
 
   static defaultProps = {
-    disabled: false,
     publicKey: null,
     deviceInfo: null,
     deviceError: null,
-    onLogin: noop
+    onLogin: noop,
+    progress: 'INITIAL'
```
mhuggins: Let's not use a string value here.  All the constants we need are defined in `progressValues`.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -56,7 +57,8 @@ export default class LoginFormLedger extends React.Component {
   }
 
   renderActions = () => {
-    const disabled = this.props.disabled || !this.props.deviceInfo;
+    const { publicKey, deviceInfo, progress } = this.props;
+    const disabled = !(publicKey && deviceInfo && (progress !== FAILED));
```
mhuggins: I think we can simplify this just by looking at `progress` and `disabled` alone:

```js
const disabled = this.props.disabled || this.props.progress !== LOADED;
```

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -66,36 +68,20 @@ export default class LoginFormLedger extends React.Component {
...
   }
 
-  renderDeviceInfo = () => {
-    const { deviceInfo } = this.props;
+  renderStatus = () => {
+    const { publicKey, deviceError, deviceInfo, progress } = this.props;
 
-    if (deviceInfo) {
-      return <p>Found USB {deviceInfo.manufacturer} {deviceInfo.product}</p>;
+    if (deviceError !== 'No USB device found.') {
```
mhuggins: By wrapping the error check as the outer condition, this means that if an error happens once followed by success, I believe we'll always end up in the `else` condition where we show the "Searching for USB devices" messaging.

src/login/components/LoginFormLedger/LoginFormLedger.js
```diff
@@ -66,36 +68,20 @@ export default class LoginFormLedger extends React.Component {
...
+      if (publicKey && deviceInfo && (progress !== FAILED)) {
```
mhuggins: In conjunction with my previous comment, I don't think we should consider `publicKey` or `deviceInfo` anymore for conditionals now that `progress` is available.  We should just use `progress` alone to determine what messaging to show.

* If it's `LOADED`, then we know that `deviceInfo` & `publicKey` are available.
* If it's `FAILED`, then we know that `deviceError` is available.
* If it's neither of the above, then we know we're searching for USB device.

tjuretzek: Regarding the three comments above: I know that it looks a bit wonky at the moment. But the problem here is, that the progress changes it's state every second (when the inverval function refreshes the props). Therefore (if you output it to the console) you get a "LOADING", "LOADED", "LOADING", "LOADED", ... or a "LOADING", "FAILED", "LOADING", "FAILED", ... loop.

A simple progress === LOADED or progress === FAILED doesn't work at the moment, because of the LOADING steps between.

I thought about mapping LOADING and FAILED to handle them as false and only LOADED as true, but I failed with the result of a flickering deviceError message or a flickering connected to Ledger message.

I'll try to find a solution for that.

mhuggins: Hmm, the `recentlyCompletedStrategy` is supposed to make it so we only see "LOADING" a single time, and once it progress to "LOADED" or "FAILED" it will never regress.  Is your experience seeing that before or after adding `recentlyCompletedStrategy`?

mhuggins: For reference: https://github.com/neoverse/spunky#recently-completed-strategy

tjuretzek: I'm seeing this behavior on the current implementation.
Everytime the setInterval(this.props.poll, POLL_FREQUENCY);  executes, the progress returns 'LOADING' and some milliseconds later 'FAILED' or 'LOADED'.

I refactored the code to match your comments above and this is the last piece missing. Sadly the deviceError stays as it is even when the process changes to 'LOADED', otherwise I would have used it as a condition for success.

I could commit my current state of work to the branch, maybe I'm overseeing something trivial?

mhuggins:
Thanks for the update, I'll take a closer look when I'm in front of my
computer again in a couple hours.

src/login/components/LoginFormLedger/LoginFormLedger.js
tjuretzek: @mhuggins I know you are not a fan of the state approach, but as long as recentlyCompletedStrategy returns the LOADING state, would the solution I provided in the last push feasible (working with componentDidUpdate and ignoring the LOADING states of progress)?

tjuretzek:
@mhuggins were you able to review the current solution?

src/login/components/LoginFormLedger/index.js
```diff
@@ -23,6 +23,9 @@ export default compose(
   withActions(ledgerActions, mapLedgerActionsToProps),
   withData(ledgerActions, mapLedgerDataToProps),
   withError(ledgerActions, mapLedgerErrorToProps),
+  withProgress(ledgerActions, {
+    progress: recentlyCompletedStrategy
```
mhuggins: Here's the core issue.  The key should be `strategy`, not `progress`.  I've got some changes I'll push up and open a PR against your branch.

mhuggins:
See https://github.com/tjuretzek/client/pull/1.  Once that is merged, I think this is ready.

src/login/components/LoginFormLedger/index.js
tjuretzek: Brilliant, thank you Matt! 
Sorry for wasting your time before.
You even wrote strategy in your change request after the first review, but I missed it...

mhuggins: No problem!

mhuggins:
@tjuretzek any update on this?

mhuggins:
If it's easier, we can also close this PR, and instead just open one from the [nos/ledger-login-fixes](https://github.com/nos/client/tree/ledger-login-fixes) branch.  It still includes your commits.

tjuretzek:
> See tjuretzek#1. Once that is merged, I think this is ready.

@mhuggins I thought you would merge it 🙈 
Sorry again a misunderstanding - we should stop having them :D

I'm merging your changes into my branch, so we should be good to go!

mhuggins:
No worries! I don't have privileges to merge into your fork, so I couldn't. I'll take another look when I'm in front of my computer again, thanks!

Issue #192, opened 2018-05-31
New tab should get focus on url field


robmoorman:
## Expected Behavior
If the user presses 'New Tab' there are used to directly enter the address they want.

## Current Behavior
After pressing 'New Tab' there is no focus.

## Your Environment
Every platform has the same behavior currently.

nasuru1:
You can use CTRL+L

nasuru1:
Unfortunately, they consider exposing that functionality a security risk. The browser randomly focuses or doesn't.

robmoorman:
Hi @nasuru1 in which way this is a security risk, can you please elaborate?

Besides the CTRL+L this still applies; by opening a new tab the end user expects to directly enter a address without clicking first or hitting ctrl+l.

PR #194, opened 2018-05-31
feat(tabs): focus address field after adding new tab

robmoorman:
## Description
This pull requests makes sure the address field is focussed after adding a new tab to quickly enter your preferred url.

## Motivation and Context
Entering an url takes another manual click in the client; expected default browser behavior after adding a new tab is that the address field is focussed (safari, chrome, etc.).

## How Has This Been Tested?
On OSX with multiple node environments.
Tested with new tabs, closing tabs, re-new tab, etc. (see .gif).

tab-focus

## Types of changes

## Checklist:
- [ ] The commit message follows our guidelines

## Closing issues
#192

### DIFF
--- a/src/browser/components/AddressBar/AddressBar.js
+++ b/src/browser/components/AddressBar/AddressBar.js
@@ -26,6 +26,12 @@ export default class AddressBar extends React.Component {
       this.input.value = this.props.query;
       this.input.blur();
     }
+
+    if (this.props.tabCount > prevProps.tabCount) {
+      // Focus the address field after opening a new tab
+      this.input.value = '';
+      this.input.focus();
+    }
   }
 
   render() {
--- a/src/browser/components/AddressBar/AddressBar.js
+++ b/src/browser/components/AddressBar/AddressBar.js
@@ -31,7 +31,12 @@ export default class AddressBar extends React.Component {
 
     if (this.props.tabCount > prevProps.tabCount) {
       // Focus the address field after opening a new tab
-      this.input.focus();
+      this.input.select();
+    } else if (this.props.tabCount < prevProps.tabCount) {
+      // Blur the address field after closing a tab and make sure the
+      // correspondent address is shown
+      this.input.value = this.props.query;
+      this.input.blur();
     }
   }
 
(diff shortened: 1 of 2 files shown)


mhuggins:
I like the idea of focusing on the address bar immediately, but I'm not sure we want to delete the address that opens on a new tab, at least not while we're still developing and trying to drive users to the welcome screen (and make it clear that there's an `nos:` protocol).

I think we should reopen this as an issue that needs discussion related to design/UX.

jeroenptrs:
@mhuggins @robmoorman best of both worlds solution would be to open the new tab with the welcome screen address selected. But in all fairness, `chrome://newtab` isn't visible either.

robmoorman:
He @jeroenptrs, good suggestion. Although Chrome doesn't provide this either; think this is indeed best of words. Will update the PR today. Thanks.

robmoorman:
tab-focus-nos-protocol

mhuggins:
@robmoorman Would it be possible to make it focus + highlight everything in the field?

robmoorman:
Ok selected/highlighted the address now. Also blurred and reset the address while closing tabs.

tab-focus-nos-protocol

src/browser/components/AddressBar/AddressBar.js
```diff
@@ -26,6 +28,16 @@ export default class AddressBar extends React.Component {
...
+
+    if (this.props.tabCount > prevProps.tabCount) {
+      // Focus the address field after opening a new tab
+      this.input.select();
+    } else if (this.props.tabCount < prevProps.tabCount) {
+      // Blur the address field after closing a tab and make sure the
+      // correspondent address is shown
+      this.input.value = this.props.query;
+      this.input.blur();
+    }
```
mhuggins: I'm sorry for the pushback on this.  I've been thinking about this a bit since we picked our design competition winner.  With the new design, we're going to be inverting the tabs and address bar, such that the tabs appear on top, and the address bar on bottom (similar to Chrome).  What this means is that we would probably end up rendering one `AddressBar` component per `DAppContainer` (similar to the `webview` element & new `Error` component you added).  When this happens, I think it will simplify things quite a bit, such that we can simply use `componentDidMount` for this behavior.

Given that, I'd rather punt a little bit on this, until after we flip the tabs & address bar around.

robmoorman:
He @mhuggins sure thing! Will pick this up by tomorrow 👍

mhuggins:
@robmoorman I'm going to close this for now just for managerial purposes over our outstanding PR's.  Please feel free to reopen when ready!

Issue #187, opened 2018-05-28
Prevent browser tabs to grow passed application width


DalderupMaurice:
## Expected Behavior

Not sure yet - cap amount of tabs or group them? 
Might want to discuss the expected behavior first in discord or here

## Current Behavior

The tabs grow passed the application width, enforcing people to scroll horizontally

## Possible Solution

- Cap the amount of tabs allowed
- Group tabs
(Must be discussed first) 

## Screenshot
![](https://i.imgur.com/7088mNO.png)

## Steps to Reproduce (for bugs)

1. Click 1000 times on the "New tab" button

## Context

Prevent browser tabs to grow passed application width to prevent horizontal scrolling

## Your Environment

* Version used:
* Environment name and version (e.g. node.js 9.6.1, yarn 1.6.1, npm 5.6.0):
* Operating System and version (desktop or mobile):
any

mhuggins:
I think we should consider implementing one or both of the following:

1. Set a min-width on the tabs, and once there are enough tabs opened to justify it, make the tabs scroll horizontally.  This might not be ideal for desktop users since mouse wheels don't generally scroll horizontally though.
1. Set a max number of tabs and just disable the "Add" button when we reach that limit.  For comparison, I tried opening a bunch of tabs in Chrome and found that they opened indefinitely, but I ultimately ran into memory usage issues where I would have been better off with a max limit. 😆

DalderupMaurice:
I'm not a fan of horizontal scrolling :P

mmoshi:
Looks like a duplicate of https://github.com/nos/client/issues/164, can probably be closed.

DalderupMaurice:
Actually not, the tabs are now responsive but are still able to grow beyond the application it's width @mmoshi

PR #195, opened 2018-06-01
chore(tabs): adjusted the tab styles

tjuretzek:
## Description
This is an alternative approach for the tab styles. The fix is flex-box based and has a more browser typical look. There is still no maximum count of tabs, but if you keep clicking the "new tab" button it gets currently moved out of the viewport.
By reducing the padding and therefore minimizing the minimal width of a tab it is possible to add up to 60 tabs (like in Chrome) but I felt it looked better with a little space. A limitation of tabs regarding the width of the browser would be suggestion to prevent any false feeling behaviors.

## Motivation and Context
I felt the solution provided in https://github.com/nos/client/issues/187 wasn't satisfying enough to really suit the prototype, so I wanted to create a different approach for consideration.

This fix is almost CSS only (a new background div was added) and can be added as a visual fix.

## How Has This Been Tested?
It was tested on Windows and OSX with the Electron native Chrome browser.

bildschirmfoto 2018-06-01 um 12 07 25
bildschirmfoto 2018-06-01 um 12 07 51
bildschirmfoto 2018-06-01 um 12 08 07
bildschirmfoto 2018-06-01 um 12 08 25

## Types of changes

## Checklist:

## Closing issues

Fixes #

### DIFF
--- a/src/browser/components/Tab/Tab.scss
+++ b/src/browser/components/Tab/Tab.scss
@@ -4,42 +4,50 @@
   flex-wrap: nowrap;
   align-items: center;
   position: relative;
-  border-color: #dee2e6 #dee2e6 transparent;
-  padding: 8px 12px;
-  font-size: 11.52px;
-  transition: all 0.2s ease;
-  cursor: pointer;
+  padding: 5px 35px 0 20px;
+  margin: 0 -15px 0 0;
+  height: $tab-height;
 
-  &::after {
-    content: "";
-    width: 0%;
-    height: 3px;
-    background-color: #047bf8;
+  .tabBackground {
+    height: $tab-height*1.18;
+    width: calc(100% - 10px);
     position: absolute;
-    bottom: -2px;
-    left: 50%;
-    transform: translateX(-50%);
-    transition: all 0.2s ease;
+    border-radius: 4px;
+    border: 1px solid $tab-border-color;
+    border-bottom: none;
+    bottom: -6px;
+    left: 5px;
+    box-shadow: 0 0 2px  #fff inset;
+    transform: perspective(100px) rotateX(30deg);
+    z-index: 1;
+    background-color: $tab-default-color;
   }
 
   &.active {
-    .title {
-      color: #495057;
+    z-index: 60;
+    position:relative;
+
+    .tabBackground {
+      background-color: $tab-active-color;
+      box-shadow: 0 0 2px 0 $tab-active-highlight-color inset;
     }
 
-    &::after {
-      width: 100%;
+    .title {
+      color: #495057;
     }
   }
 
   .title {
     flex: 1 1 auto;
-    max-width: 100px;
+    max-width: 250px;
+    width: calc(100% - 20px);
     color: rgba(#000, 0.3);
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
     line-height: 1.25em;
+    z-index: 2;
+    padding: 0;
   }
 
   .close {
@@ -47,24 +55,25 @@
     display: flex;
     align-items: center;
     justify-content: center;
+    text-align: center;
     width: 15px;
     height: 15px;
     padding: 0;
-    margin: 0 0 0 8px;
+    margin: 0;
     color: rgba(#000, 0.3);
     background: transparent;
     border: none;
     cursor: pointer;
-
+    z-index: 3;
+    position: absolute;
+    right: 12px;
     &:hover {
-      color: #fff;
-      background: #e45b46;
-      border-radius: 50%;
+      color: rgba(#000, 0.6);
     }
 
     .closeContent {
       flex: 0 0 auto;
-      display: inline-flex;
+      display: inline-block;
     }
   }
 
@@ -72,5 +81,6 @@
     @include spin;
     flex: 0 0 auto;
     margin-right: 8px;
+    z-index: 4;
   }
 }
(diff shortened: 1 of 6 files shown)


src/root/stylesheets/variables.scss
```diff
@@ -2,6 +2,12 @@
 
 $content-bg-color: #f5f5f5;
 $primary-text-color: #6c6c6e;
+$primary-border-color: #dee2e6;
+
+$tab-active-color: #f8f8f8;
```
mmoshi: I think it would make more sense to keep component variables local to the component. It's impossible to know which tab component are we talking about here, there two so far in the app.

tjuretzek: Hmmm... I think having a centralized variables file for the whole project does make more sense. Maybe we should implement a more specific wording or a less specific wording of the variables to enable a global use.
For future theming the solution of keeping the variables local to the component doesn't look right to me.

I'm not talking about a setting nightmare like foundation has, but a colors and sizes group of variables is something you should consider.

mmoshi: For what reason doesn't it look right? I don't really see sizings / spacings and style specifics of a single component as being suitable for global use. I can see them being potentially misused. If correctly decoupled, each component should only worry about its own styles.

Common colors, states, font stylings, spacings, form outlines etc. would be useful to all components. 

I think the challenge is to provide a set of common variables that are minimal and obvious but useful to all components.

tjuretzek: "Common colors, states, font stylings, spacings, form outlines etc. would be useful to all components." 
We are talking about $tab-active-color and such, which (renamed) are usefull in a global scope.

The sizes variables are also logical in a global scope (for example the $menu-width), because the last commit to the issue showed that a "magical number" as you called it, can cause confusion.

What I'm suggesting is to rename the tab-colors to match their color: for example add a $color-light, $color-medium and a $color-dark and then use them inside the components or reasign them to other component specific variables.

But redeclaring the menu-width variable (for example) in all needed files (menu.scss and tabs.scss) doesn't look an feel right to me.

mmoshi: bq. We are talking about $tab-active-color and such, which (renamed) are usefull in a global scope.
Exactly, the active/selected color is useful in a global scope, but it wouldn't name it specific to tabs.

bq. The sizes variables are also logical in a global scope (for example the $menu-width), 
I see the menu width as local to the authenticated layout component. No other component should need to be concerned with it if it's handled there.

bq. What I'm suggesting is to rename the tab-colors to match their color: for example add a $color-light, $color-medium and a $color-dark
Right, these component colors could be reused and defined globally.

I think what we're missing is a global spec for all these colors and states, but I guess that will come out of the design discussions that are happening around the nOS design competition.

tjuretzek: @mmoshi I agree! Would you have a look at the last commit and tell me what you think of it?

mmoshi: Sure, I can add some more comments, but perhaps it would be a good idea to check with the team whether they want to commit to this design now before we we spend time making any more adjustments. Might be better to consider a redesign of the entire browsing area (bar, tabs, etc.) before doing this to see how it would fit in.

src/browser/components/Tab/Tab.scss
```diff
@@ -1,3 +1,10 @@
+// Tab Component Colors
+$tab-active-color: #f8f8f8;
```
mmoshi: @tjuretzek looks better! would suggest the following:

* remove the comments, they seem a bit redundant
* be specific about the property we intend, so color -> background-color
* consider whether the value is likely to be reused and whether we need the variable at all
* remove the default keyword, and just attach the property directly to the component name
* list component properties first, then states, then sub-components

```
$tab-height: 38px;
$tab-background-color: #eee
$tab-active-background-color: #f8f8f8
```

src/root/components/AuthenticatedLayout/AuthenticatedLayout.scss
```diff
@@ -1,3 +1,6 @@
+// AuthenticatedLayout Sizes
+$menu-width: 70px;
```
mmoshi: would always start with the component name:

`$authenticatedlayout-menu-width`

src/browser/components/Tab/Tab.scss
tjuretzek: done and done

mhuggins:
I'll take a closer look at this today.  I realize now that the previous PR related to tab width added scrollbars to the app, which we don't want.

DalderupMaurice:
I was wondering where these came from lawl

mhuggins:
Is there a way we can do this without changing out the tab styles?  We'll be reworking the tab styles, but we're going to go through the design competition results first before making decisions related to design changes.

tjuretzek:
That would be this commit:

https://github.com/nos/client/commit/69c93ded8660908d569a5a3a4d1f5c518afad68c

I could reroll to it, if you like :)

mhuggins:
That commit looks like it just has to do with the widths.  I meant the borders and overall look of the tabs changing.

tjuretzek:
If you only want the tab behavior without the new look, then it is the right commit.
But I‘ll commit an update without the new look to be sure, as soon as I‘m in front of my computer again.

mhuggins:
I misunderstood your comment.  Where you said "reroll", I thought you were talking about reverting that commit.  That's the commit that I think we want though.

mhuggins:
Thanks guys, closing in favor of #199.

PR #200, opened 2018-06-03
chore(browser): fix browser tab reload

### LINKED ISSUE #161: Preserve webview state when switching tabs
## Expected Behavior
When switching between tabs, the previous state of the webview (page loaded, scroll position) for that tab should be used.  Additionally, when navigating between the browser and other screens (e.g.: account or settings), the state of the last visible tab should be preserved.

## Current Behavior
When switching between tabs or navigating between screens, the page is reloaded within the webview.

## Possible Solution
Each webview has a "guestinstance" or "guestInstanceId".  One possibility is to use that to determine how the web contents should display.

Another options is that there appear to be non-documented `webview.attachGuest` and `webview.detachGuest` functions.  It's probably worth looking into those to see how they work.  Check out the [brave](https://github.com/brave/browser-laptop) source as an example.

1. Open 2+ tabs when different targets (e.g.: `nos://nos.neo` and `https://google.com`).
2. Scroll each tab to the scrollbar is not aligned to the top of the page.
3. Switch between tabs and observe the state reset.

## Context
We want swapping tabs to be fluid and not interfere with user/dApp interactions.

## Your Environment

mhuggins:
## Description
This fixes the issue where switching tabs results in the webview reloading.

## Motivation and Context
Switching to and from tabs should maintain state for any given webview.  That involves the current page, the scroll position, how long the user's been on the page, JS state, etc.

Prior to this PR, switching tabs results in react unmounting and remounting webview elements, which results in the page being removed from the DOM and then re-added and re-fetched.  With this change, the elements are instead just being hidden from view when the tab changes and displayed again when the tab is opened.

## How Has This Been Tested?
I smoke tested in the electron app and saw that scroll position was being maintained & no extra requests were being made.  I also added a simple set of react container tests to ensure that we're rendering all the webviews when changing tabs.

change-tabs

## Types of changes

## Checklist:

## Closing issues
Fixes #161


__tests__/browser/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+  wrapper.update();
+};
+
+describe('<Browser />', () => {
+  it('renders all tabs', () => {
+    const wrapper = mountBrowser();
+    expect(wrapper.find(DAppContainer).length).toBe(2);
+  });
+
+  it("only only marks the activeSessionId's dapp as active", () => {
```
DalderupMaurice: irrelevant to the code - but there's a duplicate word here

src/browser/components/DAppContainer/DAppContainer.scss
```diff
@@ -1,3 +1,8 @@
 .dAppContainer {
   height: 100%;
+  display: none;
```
jeroenptrs: This isn't really a fix to the reload, right? How does a regular browser handle this?

DalderupMaurice:
Before merging - consider @jeroenptrs his comment
Wouldn't this clutter the RAM when you have 20 tabs open? :P

src/browser/components/DAppContainer/DAppContainer.scss
mhuggins: It is a fix because we're now just hiding the DOM element instead of allowing React to unmount its host component.

mhuggins:
@DalderupMaurice @jeroenptrs Here's a bit of insight around webviews that will hopefully better explain the solution here.

* When a webview is added to the DOM, it is assigned a `guestinstance`, which is available as an attribute on the element.  Each new webview is assigned a unique `guestinstance` (unless it is provided one by the developer).
* The `guestinstance` is associated with `webContents`, which is essentially the context of any window or webview that shares that `guestinstance` value.  By context I mean the current page, JS state, scroll position, etc.
* When all webview elements sharing the same `guestinstance` are removed from the DOM, that `webContents` context is removed completely, and trying to assign the same `guestinstance` to another webview will not work (it behaves like you didn't ever try to assign a `guestinstance` value).  Because of this, if we want to preserve the context across tabs, we need to hide the webviews rather than removing them from the DOM.

Regarding the question about cluttering RAM, I did a bit of looking, and electron disables [offscreen rendering](https://github.com/electron/electron/blob/master/docs/tutorial/offscreen-rendering.md) by default.  And according to the docs:

> The offscreen rendering in Electron uses a similar approach than the [Chromium Embedded Framework](https://bitbucket.org/chromiumembedded/cef) project.

With all that said, I think we're safe and have to use this approach.

DalderupMaurice:
Good with me
Wbu? @jeroenptrs

Issue #137, opened 2018-05-10
Unit - integration tests of the nOS API


DalderupMaurice:
## Context
We should start with implementing unit/integration tests for the nOS API as it's essential to track it's functioning

mhuggins:
@DalderupMaurice The tests are working for me locally both on `master` and `develop`.  They're also passing on CircleCI.  Can you add more detail about what you're experiencing?

DalderupMaurice:
Ahh well, it's more like "Add tests for the nOS API" 
The command works - but the tests for the API do not exist at this moment 😄 @mhuggins

mhuggins:
Ahh, I misunderstood the description.  Thanks 😄

jeroenptrs:
@mhuggins @DalderupMaurice can we split Team and Community up in respectively setting up e2e testing environment and adding testcases?

PR #207, opened 2018-06-07
chore(tests): added tests for GetBalance & GetAddress requests

mhuggins:
## Description
This adds a thorough set of unit tests around the `GetBalance` and `GetAddress` components & containers.

## Motivation and Context
We need to have trust that what we're building works as expected, and also that we don't break it as we continue working on it.

## How Has This Been Tested?
I ran the test suite!

screen shot 2018-06-07 at 11 20 45 am

## Types of changes

## Checklist:

## Closing issues
Contributes towards #137, but doesn't close it yet.

### DIFF
--- a/__tests__/testHelpers.js
+++ b/__tests__/testHelpers.js
@@ -1,13 +1,18 @@
 import React from 'react';
 import thunk from 'redux-thunk';
 import configureStore from 'redux-mock-store';
+import createSagaMiddleware from 'redux-saga';
 import { createStore as createReduxStore, applyMiddleware } from 'redux';
 import { Provider } from 'react-redux';
+import { saga, progressValues } from 'spunky';
 
 import reducers from 'root/reducers';
 
 export const createStore = (initialState = {}) => {
-  return createReduxStore(reducers, initialState, applyMiddleware(thunk));
+  const sagaMiddleware = createSagaMiddleware();
+  const store = createReduxStore(reducers, initialState, applyMiddleware(sagaMiddleware, thunk));
+  sagaMiddleware.run(saga);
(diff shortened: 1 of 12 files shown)


src/browser/actions/makeBalancesActions.js
```diff
@@ -6,10 +6,10 @@ import generateDAppActionId from './generateDAppActionId';
 
 export const ID = 'balances';
 
-export default function makeBalancesActions(sessionId, requestId) {
+export default function makeBalancesActions(sessionId, requestId, call = getBalances) {
```
mhuggins: Trying to use jest to stub this `getBalances` default export proved pretty challenging, and it felt like a good opportunity to just use dependency injection instead.  I decided to go that route instead, and I think we should do the same for all our `makeXxxActions` that are used for the nOS API.

__tests__/testHelpers.js
```diff
@@ -20,3 +25,34 @@ export const provideState = (node, initialState = {}) => {
...
+  data
+});
+
+export const mockSpunkyLoading = ({ loadedCount = 0 } = {}) => ({
+  ...spunkyInitialState,
+  progress: progressValues.LOADING,
+  loadedCount
+});
+
+export const mockSpunkyFailed = (error, { loadedCount = 0 } = {}) => ({
```
mhuggins: My plan is to eventually move these into the spunky package, but I want to make sure they are sufficient for our use cases going forward first.

mhuggins:
I added tests for the GetBalance component itself (in addition to the container component).

screen shot 2018-06-07 at 11 46 03 am

PR #241, opened 2018-06-17
feat(encrypt): add functions to allow encryption and decryption of arbitrary data

trueinsider:
## Description
There is three new API functions:
`getPublicKey` -- provides the public key of the currently authenticated account
`encrypt` -- allows to encrypt arbitrary data
`decrypt` -- allows to decrypt previously encrypted data

## Motivation and Context
To securely exchange messages in the decentralized applications context some kind of encryption is necessary.

## How Has This Been Tested?
Tested in client connected to nos-local with the help of slightly modified template dApp. Also there is new tests added.

## Types of changes

## Checklist:
- [x] Docs have been added / updated (for bug fixes / features)


__tests__/browser/components/RequestsProcessor/Decrypt/index.test.js
```diff
@@ -0,0 +1,121 @@
...
+      expect(wrapper.find('NullLoader').length).toBe(1);
+    });
+
+    it('does not resolve or reject', () => {
+      expect(onResolve).not.toHaveBeenCalled();
+      expect(onReject).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('when decrypted data has loaded', () => {
```
mhuggins: All these tests show the happy path, which is great.  Can you add a test for when decryption fails to ensure that `onReject` is called and `onResolve` is not called?

__tests__/browser/components/RequestsProcessor/Encrypt/index.test.js
```diff
@@ -0,0 +1,122 @@
...
+      expect(wrapper.find('NullLoader').length).toBe(1);
+    });
+
+    it('does not resolve or reject', () => {
+      expect(onResolve).not.toHaveBeenCalled();
+      expect(onReject).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('when encrypted data has loaded', () => {
```
mhuggins: Same here, can a test be added for the unhappy path?

mhuggins:
Did a little digging on the electron side, and [buffers may actually be supported](https://github.com/electron/electron/blob/4cfe5ecaa45862a8bc927903cff9716b5a2dd072/lib/browser/rpc-server.js#L66-L67).  I haven't tested yet.  Also, I think the documentation you added for the return value is still wrong because it's returning a string, not a buffer.

docs/api.md
```diff
@@ -265,6 +265,73 @@ nos.claimGas()
...
+  .then((publicKey) => alert(`Public Key: ${publicKey}`))
+  .catch((err) => alert(`Error: ${err.message}`));
+```
+
+### `encrypt`
+The `encrypt` function allows you to encrypt arbitrary data for another user (you will need his public key, please see `getPublicKey`). It does not require the user to grant permission.
+
+#### Parameters
+* `recipientPublicKey` **string** - The public key of the recipient account.
+* `data` **string** | **Buffer** - The data to encrypt.
```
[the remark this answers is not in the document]
trueinsider: 1. fixed
2. I believe this is not a concern anymore as electron seems to support Buffers

docs/api.md
```diff
@@ -265,6 +265,73 @@ nos.claimGas()
...
+```
+
+### `decrypt`
+The `decrypt` function allows you to decrypt previously encrypted data for this user (you will need public key of the sender account, please see `getPublicKey`). It does not require the user to grant permission.
+
+#### Parameters
+* `senderPublicKey` **string** - The public key of the sender account.
+* `iv` **string** - The IV received during encryption.
+* `mac` **string** - The MAC received during encryption.
+* `data` **string** | **Buffer** - The data to decrypt.
```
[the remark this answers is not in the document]
trueinsider: I believe this is not a concern anymore as electron seems to support Buffers

docs/api.md
```diff
@@ -265,6 +265,73 @@ nos.claimGas()
...
+
+#### Returns
+**Buffer** - decrypted data
```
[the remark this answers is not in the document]
trueinsider: I believe this is not a concern anymore as electron seems to support Buffers

src/browser/actions/makeEncryptActions.js
```diff
@@ -0,0 +1,48 @@
...
+  const iv = randomBytes(16);
+  const encryptionKey = hash.slice(0, 32);
+  const macKey = hash.slice(32);
+  const encrypted = aes256CbcEncrypt(iv, encryptionKey, data);
+  const dataToMac = Buffer.concat([iv, senderPublicKey, encrypted]);
+  const mac = createHmac('sha256', macKey).update(dataToMac).digest();
+  return {
+    iv: iv.toString('hex'),
+    mac: mac.toString('hex'),
+    data: encrypted.toString('hex')
```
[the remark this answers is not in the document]
trueinsider: documentation for result doesn't mention Buffer anywhere

src/browser/actions/makeDecryptActions.js
```diff
@@ -0,0 +1,49 @@
...
+
+function aes256CbcDecrypt(iv, key, data) {
+  const cipher = createDecipheriv('aes-256-cbc', key, iv);
+  const firstChunk = cipher.update(data);
+  const secondChunk = cipher.final();
+  return Buffer.concat([firstChunk, secondChunk]);
+}
+
+const decrypt = ({ senderPublicKey, wif, iv, mac, data }) => {
+  const ecdh = ECDH('prime256v1');
```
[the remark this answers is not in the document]
trueinsider: 1. no particular reason, it's good enough
2. prime256v1 = secp256r1 this is what used by NEO itself according to this: https://github.com/neo-project/neo/blob/master/neo/Wallets/KeyPair.cs#L26

docs/api.md
```diff
@@ -265,6 +265,73 @@ nos.claimGas()
...
+nos.getPublicKey()
+  .then((publicKey) => alert(`Public Key: ${publicKey}`))
+  .catch((err) => alert(`Error: ${err.message}`));
+```
+
+### `encrypt`
+The `encrypt` function allows you to encrypt arbitrary data for another user (you will need his public key, please see `getPublicKey`). It does not require the user to grant permission.
+
+#### Parameters
+* `recipientPublicKey` **string** - The public key of the recipient account.
```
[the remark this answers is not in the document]
trueinsider: I called it recipientPublicKey to clarify that it's not your public key that you need to pass to the function.
The way ECIES/ECDH work is that when encrypting I take my private key and your public key
and when decrypting you take your private key and my public key.
That way we'll get same shared secret without exchanging our private keys.

If you want to just encrypt some data for yourself you don't need shared secret or public key, you can just use hash of your private key as encryption key. Of course, then you will need private key for decryption too.

docs/api.md
```diff
@@ -265,6 +265,73 @@ nos.claimGas()
...
+nos.encrypt({ recipientPublicKey, data })
+  .then(({ iv, mac, data }) => alert(`iv: ${iv}\nmac: ${mac}\ndata: ${data}`))
+  .catch((err) => alert(`Error: ${err.message}`));
+```
+
+### `decrypt`
+The `decrypt` function allows you to decrypt previously encrypted data for this user (you will need public key of the sender account, please see `getPublicKey`). It does not require the user to grant permission.
+
+#### Parameters
+* `senderPublicKey` **string** - The public key of the sender account.
```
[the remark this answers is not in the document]
trueinsider: Answered above

__tests__/browser/components/RequestsProcessor/Decrypt/index.test.js
trueinsider: fixed

__tests__/browser/components/RequestsProcessor/Encrypt/index.test.js
trueinsider: fixed

trueinsider:
I believe I've fixed all points of concern.
Also there is no Buffers in documentation for return values. But input parameters can take string or Buffer.

docs/api.md
```diff
@@ -212,6 +215,76 @@ nos.send({ asset: GAS, amount, receiver })
...
+
+#### Parameters
+* `config` **object** - The config options to perform this operation.
+* `config.publicKey` **string** - The public key of the sender account.
+* `config.iv` **string** - The IV received during encryption.
+* `config.mac` **string** - The MAC received during encryption.
+* `config.data` **string** | **Buffer** - The data to decrypt.
+
+#### Returns
+**Buffer** - decrypted data
```
mhuggins: @trueinsider This is the return value of type `Buffer` I was referring to.

trueinsider: oh, you right. but it actually returns Buffer.

inside decrypt:
>  return aes256CbcDecrypt(ivBuffer, encryptionKey, dataBuffer);

and inside aes256CbcDecrypt:
> return Buffer.concat([firstChunk, secondChunk]);

trueinsider:
I think maybe I should rename these functions to encryptFor/decryptFrom and then make another pair to just encrypt/decrypt stuff for myself using my private key
Or change API in a way that if no public key is provided in the args then encryption will be using only private key

trueinsider:
I was thinking, maybe I should do only getSharedSecret(publicKey) which will generate shared secret from provided public key and user's private key. That way we won't compromise private key and at the same time dApp developer could implement encryption/decryption that he wants/needs, not only hardcoded aes-256-cbc. And I can add encrypt/decrypt example in documentation on how to implement it using getSharedSecret() and aes-256-cbc.
And for encryption of my own data I could implement getPrivateSecret() which will return sha512 hash of user's private key.

mhuggins:
@trueinsider I'm communicating with the team regarding your last question here.  I'll circle back with you and let you know what we settle on.

DalderupMaurice:
Had a small discussing concluding that for the moment being - the additional feature is a too specific use-case. It's also no high risk of implementing it on a later stage, so we're happy to finalize this PR and potentially add in the additional feature later. :) 
cc FYI @mhuggins

kokahunter:
I really like the implementation. Wanted to check if there is a chance to implement this feature in the near future since the dApp neochat.neo would greatly benefit since encryption is needed for secure messaging. :)

I think to implement this feature inside of the nOS client is the only secure way to handle encryption based on the user's keys.

DalderupMaurice:
cc @mhuggins ^

mhuggins:
This change looks good to me, but I'm labeling with "don't merge" until we get a more in depth security review.  At this point, no code changes are deemed needed prior to merging.

DalderupMaurice:
0.3.0 pls? 🙈

mhuggins:
@DalderupMaurice @Deanpress since we're waiting on a review from @knaps, I can't guarantee it will happen before we're ready to release 0.3.0.

knaps:
@mhuggins - I think you have the wrong person! I'm not a contributor to this project

mhuggins:
Sorry about that!

src/renderer/browser/components/RequestsProcessor/Decrypt/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default function makeDecrypt(decryptActions) {
+  return compose(
+    withClean(decryptActions),
+
+    withProps(({ args }) => pick(args[0], CONFIG_KEYS)),
+
+    withData(authActions, mapAuthDataToProps),
+
+    withCall(decryptActions, ({ senderPublicKey, wif, iv, mac, data }) => ({
```
DalderupMaurice: We currently use `withInitialCall` everywhere for the API, is this something we should address before merging? @mhuggins

src/renderer/browser/util/getRandomIV.js
```diff
@@ -0,0 +1,5 @@
+import { randomBytes } from 'crypto';
```
DalderupMaurice: Also this dep doesn't seem to be included in the `package.json`/yarn.lock file

src/renderer/browser/util/getRandomIV.js
```diff
@@ -0,0 +1,5 @@
...
+
+export default function getRandomIV() {
+  return randomBytes(16);
```
DalderupMaurice: Would we maybe want the ability to choose the strength?
Having an optional parameter with 128 as default, and having` return randomBytes(strength /8);`

src/renderer/browser/util/getRandomIV.js
mhuggins: This is built into node.

src/renderer/browser/components/RequestsProcessor/Decrypt/index.js
mhuggins: That would potentially be safer here to ensure we don't inadvertently call the action twice (in case of a re-render for any reason).

DalderupMaurice:
I'll take these changes in a followup PR

Issue #12, opened 2018-03-14
NEO/GAS & NEP-5 Token Wallet Tab


deanpress:
Requires all basic wallet functions:
* [x] View NEO/GAS Funds
* [x] Make NEO/GAS Transactions
* [x] View NEP-5 Funds
* [ ] Make NEP-5 Transactions

jeroenptrs:
@DalderupMaurice @mhuggins I'm going to make this a `team issue`. Especially since this is a new window and seems a bit more complex than at first glance

mhuggins:
@jeroenptrs Sounds good to me 👍

DalderupMaurice:
This is related to my open PR #113 @jeroenptrs @mhuggins 
It's almost done 😂

jeroenptrs:
Before we actually start transferring NEP-5 tokens, a great step in-between might be to start listing them? @DalderupMaurice @mhuggins @deanpress

PR #247, opened 2018-06-18
feat(tokens): add token balances on account page

trueinsider:
## Description
Add ability to see NEP-5 token balances on Account page

## Motivation and Context
This is a basic wallet function requested in #12

## How Has This Been Tested?
Tested in client

tokens

## Types of changes

## Checklist:

## Closing issues
Partly implements #12


src/renderer/shared/values/tokens.js
```diff
@@ -0,0 +1,137 @@
...
+export const QLC = '0d821bd7b6d53f5c2b40e217c6defc8bbe896cf5';
+export const RHT = '2328008e6f6c7bd157a342e789389eb034d9cbc4';
+export const TKY = '132947096727c84c7f9e076c90f08fec3bc17f18';
+export const RPX = 'ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9';
+export const SOUL = 'ed07cffad18f1308db51920d99a2af60ac66a7b3';
+export const SWTH = 'ab38352559b8b203bde5fddfa0b07d8b2525e132';
+export const THOR = '67a5086bac196b67d5fd20745b0dc9db4d2930ed';
+export const TNC = '08e8c4400f1af2c20c28e0018f29535eb85d15b6';
+export const XQTA = '6eca2c4bd2b3ed97b2aa41b26128a40ce2bd8d1a';
+export const ZPT = 'ac116d4b8d4ca55e6b6d4ecce2192039b51cccc5';
```
mhuggins: Hardcoding these takes us down a path that makes use dependent upon releases every time there's a new NEP5 token, and I don't want us to have to worry about that as Neon Wallet has had to.  There's a [neo-tokens repo](https://github.com/CityOfZion/neo-tokens) that we can use to fetch the list of tokens whenever the app runs, and use that instead.  I would like for us to do that so that there is never any manual updating required in here for tokens.

Along with such a change, we won't ever need this constants file.

src/renderer/shared/util/getBalances.js
```diff
@@ -37,10 +45,20 @@ export default async function getBalances({ net, address, tokens }) {
...
+
+    delete assets.NEO;
+    delete assets.GAS;
+
+    Object.keys(assets).forEach((tokenName) => {
+      const tokenData = getTokenData(assets, tokenName);
+      balances[tokenData.scriptHash] = tokenData;
+    });
+
+    return balances;
```
mhuggins: There's another promise that comes before this one for fetching token balances.  If neoscan returns the token balances, then we can potentially drop the individual RPC calls for all token balances.  That comes at the tradeoff of trusting that neoscan is accurate, which Luciano (from CoZ) has said it is not.

Because it's not, we should probably prefer to keep the RPC request promise above for each token instead of getting this data from the API.

src/renderer/shared/values/tokens.js
trueinsider: what would be best way to use that neo-tokens package? as far as I can see it's not on yarn/npm

mhuggins: We can just fetch the JSON file directly.  https://raw.githubusercontent.com/CityOfZion/neo-tokens/master/tokenList.json

trueinsider: fixed

src/renderer/shared/util/getBalances.js
trueinsider: fixed

src/renderer/account/components/AccountPanel/index.js
```diff
@@ -10,6 +10,25 @@ import withNetworkData from 'shared/hocs/withNetworkData';
 
 import AccountPanel from './AccountPanel';
 
+const fetch = require('isomorphic-fetch');
```
mhuggins: Can we just use `node-fetch` here instead of adding another dependency to package.json?

src/renderer/account/components/AccountPanel/index.js
```diff
@@ -10,6 +10,25 @@ import withNetworkData from 'shared/hocs/withNetworkData';
...
+
+const tokenListUrl = 'https://raw.githubusercontent.com/CityOfZion/neo-tokens/master/tokenList.json';
+
+let tokens = [];
+fetch(tokenListUrl).then((data) => data.json()).then((data) => {
```
mhuggins: I'd prefer if we stick with async/await syntax within the app for consistency.  We can move this into a function to do so, e.g.:

```js
const tokens = (async () => {
  const result = await fetch(tokenListUrl);
  const data = result.json();
  // etc;
})();
```

src/renderer/account/components/AccountPanel/index.js
```diff
@@ -10,6 +10,25 @@ import withNetworkData from 'shared/hocs/withNetworkData';
...
+  tokens = Object.values(data).sort((a, b) => {
```
mhuggins: lodash will help here:

```js
import { sortBy } from 'lodash';

tokens = sortBy(data, 'symbol');
```

src/renderer/account/components/AccountPanel/index.js
trueinsider: thank you, I didn't noticed node-fetch

src/renderer/account/components/AccountPanel/index.js
trueinsider: thank you

src/renderer/account/components/AccountPanel/index.js
trueinsider: then tokens is not an array anymore but a Promise. how I should pass it's value to `withInitialCall` then? I'm not able to do `await` there.

src/renderer/account/components/AccountPanel/AccountPanel.js
```diff
@@ -20,7 +20,16 @@ export default class AccountPanel extends React.Component {
       <Panel className={styles.accountPanel} renderHeader={this.renderHeader}>
         <div className={styles.content}>
           NEO: {this.getBalance(NEO)}<br />
-          GAS: {this.getBalance(GAS)}
+          GAS: {this.getBalance(GAS)}<br />
+          {
+            Object.values(this.props.balances).map(({ symbol, balance }) => {
+              if (symbol === undefined || balance === '0') {
```
mhuggins: I think it would make more sense to filter these out inside of `AccountPanel/index.js`.  We could use the `withProps` HOC from recompose.  This component should just be pretty dumb and render whatever it's given.

src/renderer/account/components/AccountPanel/index.js
```diff
@@ -8,8 +10,33 @@ import authActions from 'login/actions/authActions';
...
+})();
+
+const tokensPromise = (async () => {
+  const currentNetwork = await getStorage(currentNetworkActionsID);
+  if (currentNetwork !== MAINNET) {
+    return [];
+  }
+
+  return tokenListPromise;
+})();
```
mhuggins: I've thought about this a bit more, and I'm concerned that if this call fails for any reason, then the entire app will die.  That's not a very good user experience.

With that in mind, I think it would be better for us to encapsulate this into a new redux/spunky action.  We would just need to ensure that call succeeds before we ever try to fetch token balance.

I can help with this if needed.

mhuggins:
Very happy to help on the change I mentioned, so ping me if interested!

src/renderer/account/components/AccountPanel/index.js
trueinsider: I've added small change so that at least NEO/GAS will still show even if tokenList fetch wasn't successful.

mhuggins:
@trueinsider Sorry for the delay on this!  I took a much closer look at this on my system.  Everything worked well, but I had a few pieces of cleanup that I wanted to make.  I basically took your token fetch code and moved that into the `balancesActions` file, and put it into a function for clarity.  I also removed the duplication of the `balances` and `tokenBalances` props, and made it so that everything is shared in a single `balances` prop.

As soon as tests pass on this, I'm ready to merge!

src/renderer/account/components/AccountPanel/AccountPanel.js
```diff
@@ -30,6 +29,18 @@ export default class AccountPanel extends React.Component {
     return `Address ${this.props.address}`;
   }
 
+  renderBalances = () => {
+    const balances = values(this.props.balances);
+
+    if (balances.length === 0) {
+      return 'No available assets.';
```
mhuggins: I also added this when there are no assets.  That way we don't have to explicitly output NEO and GAS before tokens.

PR #267, opened 2018-06-24
Updated eslint and associated packages

mhuggins:
## Description
This cherry-picks the eslint-related greenkeeper PR's into a single PR where the overall changes can be seen in unison.

## Motivation and Context
Greenkeeper had various eslint-related package updates, but it was hard to see the effect of each individual within our client codebase.  This helps to understand the impact of the change so that we can override rules and update syntax as necessary.

## How Has This Been Tested?
`yarn lint`

## Types of changes

## Checklist:

## Closing issues
N/A

### DIFF
--- a/src/renderer/shared/components/Forms/Button/Button.js
+++ b/src/renderer/shared/components/Forms/Button/Button.js
@@ -6,7 +6,7 @@ import styles from './Button.scss';
 
 export default class Button extends React.Component {
   render() {
-    return (
+    return ( // eslint-disable-next-line react/button-has-type
       <button
         {...this.props}
         ref={this.registerRef}
--- a/src/renderer/shared/hocs/withProgressChange.js
+++ b/src/renderer/shared/hocs/withProgressChange.js
@@ -21,8 +21,8 @@ export default function withProgressChange(actions, progress, callback) {
   return (Component) => {
     class WrappedComponent extends React.Component {
       componentWillReceiveProps(nextProps) {
-        if (!progresses.includes(this.props[PROGRESS_PROP])
-            && progresses.includes(nextProps[PROGRESS_PROP])) {
+        if (!progresses.includes(this.props[PROGRESS_PROP]) &&
+            progresses.includes(nextProps[PROGRESS_PROP])) {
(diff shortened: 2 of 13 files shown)


src/renderer/shared/hocs/withProgressChange.js
```diff
@@ -21,8 +21,10 @@ export default function withProgressChange(actions, progress, callback) {
   return (Component) => {
     class WrappedComponent extends React.Component {
       componentWillReceiveProps(nextProps) {
-        if (!progresses.includes(this.props[PROGRESS_PROP])
-            && progresses.includes(nextProps[PROGRESS_PROP])) {
+        if (
+          !progresses.includes(this.props[PROGRESS_PROP]) &&
```
DalderupMaurice: I'm not a big fan of this if statement xD

mhuggins: @DalderupMaurice I reverted this, other than the `&&` placement.

src/renderer/shared/components/Forms/Button/Button.js
```diff
@@ -6,7 +6,7 @@ import styles from './Button.scss';
 
 export default class Button extends React.Component {
   render() {
-    return (
+    return ( // eslint-disable-next-line react/button-has-type
```
DalderupMaurice: could you add in the type here as well, just like the other buttons? :P

DalderupMaurice:
What rule was causing the if-else to be so weird?

mhuggins:
I just thought it looked better. 🤷‍♂️

src/renderer/shared/components/Forms/Button/Button.js
mhuggins: The type is in here implicitly.  It's part of the `props`, which we splat here with the spread operator.  If you check out the `defaultProps` further down in this file, you can see it's set to "button".

mhuggins: eslint doesn't recognize it though, which is why I added the `eslint-disable-next-line` comment here.

PR #283, opened 2018-06-27
chore(tests): added tests for GetStorage component & container

mhuggins:
## Description
This adds component & container tests for the `getStorage` API call.

## Motivation and Context
We need more confidence around our API functions to ensure they're doing what's expected.

## How Has This Been Tested?
`yarn test`

## Types of changes

## Checklist:

## Closing issues
N/A


src/renderer/browser/components/RequestsProcessor/ClaimGas/index.js
```diff
@@ -14,7 +14,7 @@ import withRejectMessage from '../../../hocs/withRejectMessage';
 const mapAuthDataToProps = ({ address, wif }) => ({ address, wif });
 const mapSendDataToProps = (txid) => ({ txid });
 
-export default function makeClaimComponent(claimActions) {
+export default function makeGetClaim(claimActions) {
```
DalderupMaurice: Did you mean "makeGasClaim" or is this name intentional? xD

src/renderer/shared/util/getStorage.js
```diff
@@ -0,0 +1,25 @@
+import { wallet, api, rpc, u } from '@cityofzion/neon-js';
+
+export const encode = (value) => {
```
DalderupMaurice: we could write this like so:
```
export const encode = value => u.str2hexstring(value);
```
Having it like this is an implicit return statement - also It's shorter and cleaner xD
(I think our linter will add parentheses around `value`, but that's okay)

src/renderer/browser/components/RequestsProcessor/ClaimGas/index.js
mhuggins: Whoops, meant for this to be `makeClaimGas`.  The format is supposed to be `make<ComponentName>`.  Thanks!

PR #289, opened 2018-06-30
feat(nep5transfer): ability to transfer NEP-5 tokens

### LINKED ISSUE #12: NEO/GAS & NEP-5 Token Wallet Tab

trueinsider:
## Description
Add ability to transfer NEP-5 tokens from Account page

## Motivation and Context
This is a basic wallet function requested in #12

## How Has This Been Tested?
Tested in client, tokens transferred: https://neotracker.io/tx/d79d2d8cc78b2a87060cc94ecd0beb91011a9cd4f5a554a8929ea2298152ab8e

## Types of changes

## Checklist:

## Closing issues
closes #12


src/renderer/shared/util/sendAsset.js
```diff
@@ -16,17 +14,38 @@ export default async function sendAsset({ net, asset, amount, receiver, address,
...
+    };
+
+    ({ response: { result, txid } } = await api.sendAsset(config, api.neoscan));
+  } else {
+    ({ response: { result, txid } } = await Neon.doInvoke({
+      net,
+      address,
+      script: createScript(asset, 'transfer', [
+        u.reverseHex(wallet.getScriptHashFromAddress(address)),
+        u.reverseHex(wallet.getScriptHashFromAddress(receiver)),
```
DalderupMaurice: normally createScript handles this [here](https://github.com/nos/client/blob/5c4d587289ab7036f486291e51c24ef65ca39ec8/src/renderer/browser/util/encodeArgs.js#L15) when encoding is enabled

Although It does not contain the parsing of the Fixed8 float - not sure how we could differentiate it in from a number/Fixed8 using the tree library from @jeroenptrs 

Maybe this is something we should consider?

jeroenptrs: Fixed8 is just a hexstring? Wouldn’t it be very difficult to discern?

DalderupMaurice: you could pass in a Fixed8 is literally a float with 8 decimals places :P

src/renderer/shared/util/sendAsset.js
```diff
@@ -33,10 +33,10 @@ export default async function sendAsset({ net, asset, amount, receiver, address,
...
       address,
       script: createScript(asset, 'transfer', [
-        u.reverseHex(wallet.getScriptHashFromAddress(address)),
-        u.reverseHex(wallet.getScriptHashFromAddress(receiver)),
-        new u.Fixed8(amount).toReverseHex()
-      ], false),
+        address,
+        receiver,
+        new u.Fixed8(amount)
+      ], true),
```
DalderupMaurice: It defaults to true if it's not passed :) 
Did you also confirm that it results in the same behavior?
LGTM though - going to let @mhuggins go through the PR for final check ^^

trueinsider: I couldn't make sure that it defaults to true looking at: https://github.com/trueinsider/client/blob/0ff2aab8dfd7b89b14d2ce40df516917c36cd641/src/renderer/shared/util/createScript.js#L5

I did confirm by making another successful transaction on mainnet: https://neotracker.io/tx/c7be1aa52aa4f36df4dc190ee8cff1caace80cffcce1ffb39bae23bfc33adeb7

DalderupMaurice: You can find it right [here](https://github.com/trueinsider/client/blob/0ff2aab8dfd7b89b14d2ce40df516917c36cd641/src/renderer/browser/actions/makeTestInvokeActions.js#L34). Maybe not the best place for it :P

DalderupMaurice: That said - [doInvoke](https://github.com/trueinsider/client/blob/0ff2aab8dfd7b89b14d2ce40df516917c36cd641/src/renderer/browser/actions/makeInvokeActions.js#L13) doesn't have it by default...
We might refactor this a little to make it consistent :P

trueinsider: Of couse I maybe wrong but I believe this doesn't apply in current case because I'm calling `createScript` function directly, not through `createActions`. And this function doesn't have default value for `encodeArgs` argument.

src/renderer/account/components/SendPanel/SendPanel.js
```diff
@@ -1,19 +1,23 @@
...
 import Input from 'shared/components/Forms/Input';
 import Icon from 'shared/components/Icon';
 import Select from 'shared/components/Forms/Select';
-import { ASSETS, NEO } from 'shared/values/assets';
+import { NEO } from 'shared/values/assets';
 
 import styles from './SendPanel.scss';
 import isNumeric from '../../util/isNumeric';
 
+import balanceShape from '../../shapes/balanceShape';
```
mhuggins: :neckbeard: No need for the extra whitespace above this line, it can be grouped with the other relative imports.

src/renderer/account/components/SendPanel/SendPanel.js
```diff
@@ -134,6 +151,10 @@ export default class AccountTxPanel extends React.Component {
     this.props.setReceiver(event.target.value);
   };
 
+  getAssets = (balances) => {
```
mhuggins: Since this argument is always `this.props.balances`, I would suggest not passing an arg, and just accessing via `this.props` directly from this function.  i.e.:

```js
getAssets = () => {
  return mapValues(keyBy(values(this.props.balances), 'scriptHash'), 'symbol');
}
```

src/renderer/account/components/SendPanel/index.js
```diff
@@ -31,6 +36,11 @@ export default compose(
 
   withData(authActions, mapAuthDataToProps),
   withNetworkData(),
+  withInitialCall(balancesActions, ({ net, address }) => ({ net, address })),
```
mhuggins: There's a stealthy issue here, and it's that both `SendPanel` and `AccountPanel` are being rendered at the same time, and both of them include `withInitialCall(balancesActions, ...)`.  As a result, we'll end up fetching balance data twice, which is already an expensive call.  Now that we have two child components that need this data, this call should be moved up one level into `account/components/index.js`.

mhuggins: I would also suggest moving the withProgressComponents HOC call up a level from account/components/AccountPanel/index.js into account/components/index.js. That way, the account screen waits for all balance data to load before trying to render.

In conjunction with that, I believe we can remove the `withLoadingProp` and `withProgressChange` HOC calls further down in this file.

src/renderer/account/components/SendPanel/index.js
```diff
@@ -31,6 +36,11 @@ export default compose(
...
+  withData(balancesActions, mapBalancesDataToProps),
+  withProps(({ balances }) => ({
+    balances: pickBy(balances, ({ balance }) => balance !== '0')
```
mhuggins: We're never using the keys from this object in the component.  I think we can just pass the values here instead (i.e.: just wrap this in a `values(...)` call).  That will also require changing the propType from `objectOf(...)` to `arrayOf(...)`.

src/renderer/browser/actions/makeInvokeActions.js
```diff
@@ -4,8 +4,9 @@ import { isArray } from 'lodash';
 
 import { GAS } from 'shared/values/assets';
 
+import createScript from 'shared/util/createScript';
```
mhuggins: :neckbeard: Whitespace not needed above this line since they're being imported from the same folder.

src/renderer/shared/util/encodeArgs.js
```diff
@@ -10,12 +10,15 @@ const encodeArgs = (args) => {
...
   const encodedArgs = args.map((arg) => {
+    const isFixed8 = (input) => input instanceof u.Fixed8;
+
     const executor = tree`
       lazy ${arg}
         expression ${wallet.isAddress} ${() => u.reverseHex(wallet.getScriptHashFromAddress(arg))}
         expression ${Array.isArray} ${() => encodeArgs(arg)}
         type ${'string'} ${() => u.str2hexstring(arg)}
         type ${'number'} ${() => u.int2hex(arg)}
+        expression ${isFixed8} => ${() => arg.toReverseHex()}
```
mhuggins: 👍

src/renderer/shared/util/sendAsset.js
```diff
@@ -16,17 +14,38 @@ export default async function sendAsset({ net, asset, amount, receiver, address,
...
+    const config = {
+      net,
+      address,
+      privateKey: new wallet.Account(wif).privateKey,
+      intents
+    };
+
+    ({ response: { result, txid } } = await api.sendAsset(config, api.neoscan));
+  } else {
+    ({ response: { result, txid } } = await Neon.doInvoke({
```
mhuggins: Can you use `api.doInvoke` here for consistency?

src/renderer/shared/util/sendAsset.js
```diff
@@ -16,17 +14,38 @@ export default async function sendAsset({ net, asset, amount, receiver, address,
...
+      net,
+      address,
+      script: createScript(asset, 'transfer', [
+        address,
+        receiver,
+        new u.Fixed8(amount)
+      ], true),
+      privateKey: wif,
+      gas: 0,
+      intents: [{
```
mhuggins: This drop of GAS shouldn't be necessary for invoking transfers anymore.  This was only needed by older versions of neon-js.

src/renderer/shared/util/sendAsset.js
```diff
@@ -16,17 +14,38 @@ export default async function sendAsset({ net, asset, amount, receiver, address,
...
+        assetId: GAS,
+        value: '0.00000001',
+        scriptHash: wallet.getScriptHashFromAddress(address)
+      }]
+    }));
```
mhuggins: This function has a bit of duplication now.  I think some of it can be prevented.  Perhaps reusing the config object is a good starting point, and maybe encapsulating the divergent code paths into a helper function would help.  e.g.:

```js
export default async function sendAsset({ net, asset, amount, receiver, address, wif }) {
  // snip...

  const send = () => {
    const config = { net, address, privateKey: wif };

    if (keys(ASSETS).includes(asset)) {
      const selectedAsset = ASSETS[asset];
      const intents = api.makeIntent({ [selectedAsset]: amount }, receiver);

      return api.sendAsset({ ...config, intents }, api.neoscan);
    } else {
      const script = createScript(asset, 'transfer', [address, receiver, new u.Fixed8(amount)], true);

      return api.doInvoke({ ...config, script, gas: 0 }, api.neoscan);
    }
  };

  const { response: { result, txid } } = await send();
 
  // snip...
}
```

src/renderer/shared/util/sendAsset.js
trueinsider: Sure, will do. I just did it like in makeInvokeActions.

src/renderer/shared/util/sendAsset.js
trueinsider: Ok, I didn't know that.

src/renderer/shared/util/sendAsset.js
mhuggins: Didn't realize we were using it there.  Would rather we use `api` there as well.  Thanks for pointing that out. 😄

trueinsider:
@mhuggins I've fixed things you've mentioned, please look again

PR #301, opened 2018-07-05
feat(nav): open pages in tabs

mhuggins:
## Description
This updates the address bar and navigation to match the designs.  It also updates the navigation links to open in tabs.

## Motivation and Context
To move towards parity with the latest designs.

## How Has This Been Tested?
Smoke tested navigating to `nos:` pages and `my:` pages.  There's still some cleanup work (e.g.: navigating to invalid `my:` pages; or clicking a nav icon then navigating to an `nos:` page), and some unanswered questions (e.g.: how to implement back/forward buttons for `my:` pages; or if this should be a registered protocol).

my-tabs

## Types of changes

## Checklist:

## Closing issues
Fixes #


mhuggins:
Another option for getting most of this work merged in would be to continue opening the sidebar links in their own routes, and hiding the address bar/navigation icons for those pages.

DalderupMaurice:
One thing I notice in the GIF is that the loading icon is showing forever

mhuggins:
@DalderupMaurice 

> One thing I notice in the GIF is that the loading icon is showing forever. (Infinite loading? xD) 
Also, Ive pulled the branch on my Windows PC and the icons do NOT show for me :P 

This is fixed with the latest changes, ignore the gif. 😄

> Another little thing is this - when many tabs are open, it smashes the minimize/maximize btn etc.

Thanks, I pushed an extra commit that fixes this.

> On hover the mouse should be a pointer and the text may not be selectable.

I intentionally began removing the "hand" pointer since this implies that something is link.  Since we're building an app, the regular mouse cursor feels more appropriate to use across the board for anything that's not a link.

> The dApps icon (browser?) does not work

Yes, this is intentional since we don't have that functionality yet.  (I can hide the icon until we're ready for that if preferred though.)

> Currently - the bar is disabled with the tabname in it right? Or are we going to move to any of these

After talking with Dean about it, we didn't want to support navigating to these pages via a protocol since it could result in a scenario where a dapp tricks a user into navigating to a fake account/settings tab.  This implementation will prevent that from happening.

Because of this decision, we're simply rendering a React component that is not controlled via a webview, and as such we won't have navigation history to track via a UUID or anything like that.

> Final note - when you have account already open, but no on the active screen and you click on account, it should not open a new tab, but switch to the already existing tab. That would be better than to create a new tab.

I agree, but I'd prefer to tackle that as a follow-on PR.

PR #349, opened 2018-07-24
feat(wallet): implement new design

mhuggins:
## Description
This implements a large chunk of the new wallet design. There are still some outstanding pieces, and some polish is needed (particularly around the donut chart), but I would prefer to tackle them as part of a follow-up PR.

This also fixes some little issues I came across in the process, such as not checking the number of decimal places for NEP5 tokens in the send panel, instead only checking if it's NEO/GAS.

## Motivation and Context
We're implementing the new designs into the client.

## How Has This Been Tested?
By using the client and writing some simple tests.

screen shot 2018-07-24 at 12 30 26 am

## Types of changes

## Checklist

#### Documentation

## Closing issues
N/A


package.json
```diff
@@ -21,6 +21,7 @@
     "electron-is-dev": "0.3.0",
     "electron-json-storage": "4.1.1",
     "es6-promisify": "6.0.0",
+    "format-currency": "^1.0.0",
```
DalderupMaurice: pin deps :P

src/renderer/account/actions/pricesActions.js
```diff
@@ -0,0 +1,28 @@
+// @flow
```
DalderupMaurice: Remove flow tag

src/renderer/account/actions/pricesActions.js
```diff
@@ -0,0 +1,28 @@
...
+  if (data.error) {
+    throw new Error(data.error);
+  }
+
+  return mapPrices(data, currency);
+}
+
+export const ID = 'prices';
+
+export default createActions(ID, ({ currency = 'usd' }) => () => getPrices(currency));
```
DalderupMaurice: Is this something which should be determined by the currency settings later on?

src/renderer/root/stylesheets/global.scss
```diff
@@ -1,7 +1,9 @@
-@import url('https://fonts.googleapis.com/css?family=Rubik:300,400,700');
+@import url('https://rsms.me/inter/inter-ui.css');
+// @import './fonts/inter-ui';
```
DalderupMaurice: remove comment

DalderupMaurice:
Looks good so far, although I've "ignore" the code regarding the donut thing.
There are also some pieces of code in comment (in scss especially)

I've ran it on my Mac with an empty account and it looks good so far (besides for the donut not showing because I don't have a balance.

1 Thing maybe worth mentioning is the proportions.
Looking at this screenshot below - the proportions/sizes look a little weird/empty on fullscreen.

image

src/renderer/account/actions/pricesActions.js
mhuggins: Yep, that's the idea!

mhuggins:
@DalderupMaurice Thanks for taking a look and for the feedback!  I addressed the few issues you pointed out.  Also, to answer your specific questions in your last comment...

> I've ran it on my Mac with an empty account and it looks good so far (besides for the donut not showing because I don't have a balance.

Yeah, I reached out on figma about getting a zero-state, so we'll just have to accept the lack of donut chart until that's ready.

> 1 Thing maybe worth mentioning is the proportions.
> Looking at this screenshot below - the proportions/sizes look a little weird/empty on fullscreen.

Agreed!  I think what we should do is add a max-width to the panel and leave space on the sides, similar to the new settings screen mockups.  I'll see what @fabric-8 thinks.

PR #364, opened 2018-07-28
Minor rendering performance improvements

### LINKED ISSUE #355: Improve component performance
Our React components as a whole stand to be optimized for improved performance.

Resizing the window is very slow to redraw and demonstrates the fact that a lot of re-rendering is happening.

* Make components "pure" where possible.  (Can use recompose's `pure` HOC for this.)
* Use `shouldComponentUpdate` to conditionally re-render on a reduced prop scope wherever possible.

mhuggins:
## Description
This is a small improvement for overall rendering performance.

## Motivation and Context
Because we're doing a good job with not modifying props and states directly, we can safely use `React.PureComponent` instead of `React.Component`.

This also debounces window resize events to prevent constant redraws for every incremental resize before drag/drop completes.

## How Has This Been Tested?
Navigating throughout the app to ensure nothing broke, and compared window resize performance.

N/A

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #355

### DIFF
--- a/src/renderer/root/components/App/ButtonBar/ButtonBar.js
+++ b/src/renderer/root/components/App/ButtonBar/ButtonBar.js
@@ -2,6 +2,7 @@ import React from 'react';
 import classNames from 'classnames';
 import { string } from 'prop-types';
 import { remote } from 'electron';
+import { debounce } from 'lodash';
 
 import Icon from 'shared/components/Icon';
 
@@ -21,12 +22,12 @@ export default class ButtonBar extends React.PureComponent {
   };
 
   componentDidMount() {
-    this.updateIsMax();
-    window.addEventListener('resize', this.updateIsMax);
+    this.handleUpdateMaximized();
+    window.addEventListener('resize', this.handleResize);
   }
 
   componentWillUnmount() {
-    window.removeEventListener('resize', this.updateIsMax);
+    window.removeEventListener('resize', this.handleResize);
   }
 
   render() {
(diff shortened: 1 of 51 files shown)


src/renderer/root/components/App/ButtonBar/ButtonBar.js
```diff
@@ -67,17 +68,19 @@ export default class ButtonBar extends React.Component {
...
 
-  showWindowIcons = () => {
-    return process.platform !== 'darwin';
-  }
-
-  updateIsMax = () => {
+  handleUpdateMaximized = () => {
     const win = this.getBrowserWindow();
     if (win) {
       this.setState({ isMaximized: win.isMaximized() });
```
DalderupMaurice: Mutating state here

mhuggins: @DalderupMaurice It's mutating state, but it's not mutating the existing state object.  We can still use `PureComponent` because of that differentiation.  If we were updating a state object/nested object directly, then it'd be an issue.

Here's an example that might help clarify.

```js
let object = { foo: 'foo' };

// bad
object.bar = 'bar';

// good
object = { ...this.object, bar: 'bar' };
```

In the first scenario, we're mutating the object, but it's still the same object, so PureComponent would not know there is a change since it just does a shallow equality check.  In the second scenario, we're creating a new object, and so it would be seen as a new object since a shallow equality check returns false.

PR #385, opened 2018-07-31
feat(account): receive tab & update tab designs

mhuggins:
## Description
This PR:
* adds the Send/Receive tabs to the account screen,
* updates the tab styles,
* updates the button styles, and
* updates the global link color.

## Motivation and Context
Working on implementing updated designs.

## How Has This Been Tested?
Smoke testing & running `yarn test`.

login

send-receive

## Types of changes

## Checklist

#### Documentation

## Closing issues
N/A


package.json
```diff
@@ -90,6 +91,7 @@
     "eslint-plugin-react": "7.10.0",
     "identity-obj-proxy": "3.0.0",
     "jest": "23.4.1",
+    "jest-canvas-mock": "^1.1.0",
```
DalderupMaurice: pin ^^
Maybe add in a `.yarnrc` file to pin deps by default?

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -45,49 +44,46 @@ export default class SendPanel extends React.PureComponent {
...
-      </Panel>
+      <form className={classNames(styles.send, className)}>
+        <Select
+          className={styles.asset}
+          id="asset"
+          label="Token to send"
+          value={asset}
+          onChange={this.handleChangeAsset}
+        >
+          {this.renderAssets()}
```
DalderupMaurice: Following the design, the default value - if no tokens are "found" should be `No tokens available`

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -45,49 +44,46 @@ export default class SendPanel extends React.PureComponent {
...
+          value={amount}
+          onChange={this.handleChangeAmount}
+        />
+        <Input
+          className={styles.recipient}
+          id="recipient"
+          type="text"
+          label="Recipient"
+          placeholder="Wallet address"
+          value={receiver}
```
DalderupMaurice: receiver - recipient 
I'd prefer 1 term to go with for consistency

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -45,49 +44,46 @@ export default class SendPanel extends React.PureComponent {
...
+          id="amount"
+          type="number"
+          label="Sending amount"
+          placeholder={`Enter ${symbol} amount`}
+          min="0"
+          step={step}
+          value={amount}
+          onChange={this.handleChangeAmount}
+        />
+        <Input
```
DalderupMaurice: This is later converted to an Input w/ button if we have the address book part, If I am right

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -45,49 +44,46 @@ export default class SendPanel extends React.PureComponent {
...
+          className={styles.recipient}
+          id="recipient"
+          type="text"
+          label="Recipient"
+          placeholder="Wallet address"
+          value={receiver}
+          disabled={false}
```
DalderupMaurice: this never changes :P can be left out I think

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -45,49 +44,46 @@ export default class SendPanel extends React.PureComponent {
...
-          >
-            Next
-          </Button>
-        </form>
-      </Panel>
+      <form className={classNames(styles.send, className)}>
+        <Select
+          className={styles.asset}
+          id="asset"
+          label="Token to send"
```
DalderupMaurice: Is token the right term? Why did we derivate from `asset`?

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
```diff
@@ -0,0 +1,67 @@
...
+
+const TAB_SEND = 'send';
+const TAB_RECEIVE = 'receive';
+
+const TABS = {
+  [TAB_SEND]: <IconTab renderIcon={SendIcon}>Send</IconTab>,
+  [TAB_RECEIVE]: <IconTab renderIcon={ReceiveIcon}>Receive</IconTab>
+};
+
+export default class TransactionsPanel extends React.PureComponent {
```
DalderupMaurice: This name is going to be confusing when you add in the 3rd tab - `Transactions` which would basically also be a `Transactionspanel`

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  render() {
+    return (
+      <Panel className={classNames(styles.transactionsPanel, this.props.className)}>
+        <Tabs
+          tabsClass={styles.tabs}
+          tabClass={styles.tab}
+          tabs={TABS}
+          selectedTab={this.state.tab}
```
DalderupMaurice: Just a question - I see that we almost never destructure, is this a rule we disabled? I thought it was a "required" thing to do with the latest version of eslint. I personally have had some benefits with always destructuring.

This should not be addressed right now though :P Just wondering about the fact and if we should/should not

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: I'd like to use recipient everywhere, but we may need to deprecate receiver in the API before we can be rid of it completely.

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: Yeah, I was planning to tackle that in another PR.  This didn't feel like the time to do it since I was trying to work on the receive tab for this PR.

package.json
mhuggins: Didn't know that was an option for pinning these when installing, I'll look into it!

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: Yeah, none of this is new, I just changed the filename.

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: Yup, dunno when this got added in here. 😛

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: We should decide what term to use for these.  NEO & GAS are assets, but all other NEP5's are tokens, is my understanding.

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
mhuggins: Yeah, I'm open to another name for this panel.  Let me know if you have any suggestions!

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
mhuggins: Are you talking about having to do this?

```js
const { tab } = this.state;
return <Tabs selectedTab={tab} />;
```

mhuggins:
@DalderupMaurice I updated the spots that were actionable, left comments otherwise.  Let me know what you think.

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
DalderupMaurice: Can't think of something so quickly, maybe `WalletInteractionPanel` or such sounds a little clompy/stupid hahaha

src/renderer/account/components/TransactionsPanel/TransactionsPanel.js
DalderupMaurice: Yes, at work I am forced to do so everywhere :P

src/renderer/account/components/TransactionsPanel/TransactionsPanel.scss
```diff
@@ -0,0 +1,12 @@
+.transactionsPanel {
+  max-width: 316px;
```
DalderupMaurice: the size looks a little odd when it's below the donut instead of next to it

mhuggins:
Fixed the responsive issue.  Let's thinking about that component naming.  Doesn't feel like a block to me. 😄

PR #430, opened 2018-08-12
chore(toast): add error & info icons

mhuggins:
## Description
This adds error & info icons for those types of toasts.

## Motivation and Context
The success icon was being used as a placeholder for all toast types prior to this change.

## How Has This Been Tested?
Smoke testing the various types, plus added test for the `withToast` HOC.

screen shot 2018-08-11 at 9 53 20 pm

## Types of changes

## Checklist

#### Documentation

## Closing issues
N/A


src/renderer/shared/hocs/withToast.js
```diff
@@ -11,11 +11,11 @@ function withToast({ type, propName = 'showToast', ...options } = {}) {
 }
 
 export function withSuccessToast({ propName = 'showSuccessToast', ...options } = {}) {
-  return withToast({ ...options, propName, type: TYPE_ERROR });
+  return withToast({ ...options, propName, type: TYPE_SUCCESS });
 }
 
-export function withErrorToast({ propName = 'showErrorToast', ...options } = {}) {
-  return withToast({ ...options, propName, type: TYPE_SUCCESS });
+export function withErrorToast({ propName = 'showErrorToast', autoDismiss = false, ...options } = {}) {
```
DalderupMaurice: Why is this defaulted to false - I sits there forever even when browsing to other pages :P

DalderupMaurice:
Another commit - unrelated to this PR
The `close` text doesn't have a :hover style & the close animation is a little too snappy

src/renderer/shared/hocs/withToast.js
mhuggins: An error message seemed important enough to leave up in case someone is multitasking so they don't miss it.  I agree that it doesn't make sense for it to remain on the screen when navigating away from a page though, I'll look into clearing flashes when the route changes.

mhuggins:
I'll ping the design team for a hover style on the "Close" text.  I think it could use one as well.

src/renderer/shared/hocs/withToast.js
mhuggins: @DalderupMaurice I added another commit here so that any open toasts are automatically close when the route changes.

mhuggins:
I added a comment to figma about the "Close" text highlighting, I'd rather do that as a follow-on PR though so these updates aren't blocked.

PR #470, opened 2018-08-26
 feat(browser): added dynamic protocol handling

### LINKED ISSUE #372: URLs entered without a protocol should only redirect to `nos://` if the TLD is registered on nOS (currently only .neo)
## Expected Behavior
When you enter a URL such as `github.com` instead of `http://github.com`, the client should redirect you to `http://github.com`. Only `.neo` domains should currently automatically redirect to `nos://` URLs.

## Current Behavior
Entering any URL without defining the protocol will auto-prepend `nos://`.

* nOS Client version used: dev latest
* Operating System and version (desktop or mobile): Windows 10

DalderupMaurice:
## Description
Automatically determines the TLD and redirect to the appropriate protocol
`.neo` --> `nos:` protocol
`localhost:` --> `http:` protocol
any other tld --> `https` enforced

**note** shall we enforce `https`?
For example: in chrome you cannot surf to `http://google.com` as it redirects you to `https` aswell

## Motivation and Context
#372 & ease of use

## How Has This Been Tested?
browsing (will add tests later)

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #372


deanpress:
@DalderupMaurice +1 for enforcing HTTPS

src/renderer/browser/values/protocols.js
```diff
@@ -0,0 +1,11 @@
+export const customProtocols = {
```
mhuggins: All our other `values` files export constants in uppercase.  Can we call these `PROTOCOLS~ and `CUSTOM_PROTOCOLS` for consistency?

src/renderer/browser/util/parseURL.js
```diff
@@ -1,7 +1,16 @@
...
-  const formattedQuery = trimmedQuery.includes('://') ? trimmedQuery : `nos://${trimmedQuery}`;
-  return new URL(formattedQuery);
+  const trimmedQuery = trimEnd(trim(query), '/').split('://').pop();
+
+  const tld = filter(customProtocols, (protocol) => endsWith(trimmedQuery, protocol.tld));
+  const tldToApply = isEmpty(tld) ? protocols.https : tld[0].protocol;
+
+  const url = trimmedQuery.includes('localhost:')
+    ? `${protocols.http}//${trimmedQuery}`
+    : `${tldToApply}//${trimmedQuery}`;
```
mhuggins: Let's always forward to http.  Any server that enforces https will automatically redirect from http.

Here's an example of how Chrome handles no protocol being specified, you can see it always leads to http.  That seems safer since we can't assume that a server implements https.

https-redirect

mhuggins: @deanpress Should we force that upon the user?  I know most sites redirect from http to https these days, but if the user types in an address without the TLD and it doesn't support https, then the user will get a 404, which might be confusing.

DalderupMaurice: Yeahh, mostly with nginx they set a permanent redirect to `https`, but inside the client, if I enforce `http` and I go to google, it stays with `http` :P 
The downside is that you won't be able to visit any website over `http` as it's enforced in the client in this PR

src/renderer/browser/values/protocols.js
DalderupMaurice: Ah right, forgot that 😄

src/renderer/browser/util/parseURL.js
mhuggins: @DalderupMaurice I think it only looks like it stays http because of a bug with the address not updating on redirects, which I'm fixing right now.

DalderupMaurice: I'll mark it as blocked until the PR is up @mhuggins

deanpress: @mhuggins Good point. Let's stick with http and not enfore https within the client.

DalderupMaurice:
`https://status.circleci.com/` minor outage, build is failing because of that <--

DalderupMaurice:
[](https://gyazo.com/7f90937ea17696117170a2dd511be5ae.gif)
New gif with the new changes

src/renderer/browser/util/parseURL.js
```diff
@@ -1,7 +1,12 @@
...
+import { PROTOCOLS, CUSTOM_PROTOCOLS } from '../values/protocols';
 
 export default function parseURL(query) {
-  const trimmedQuery = trim(query);
-  const formattedQuery = trimmedQuery.includes('://') ? trimmedQuery : `nos://${trimmedQuery}`;
-  return new URL(formattedQuery);
+  const trimmedQuery = trimEnd(trim(query), '/').split('://').pop();
+
+  const tld = filter(CUSTOM_PROTOCOLS, (protocol) => endsWith(trimmedQuery, protocol.tld));
+  const tldToApply = isEmpty(tld) ? PROTOCOLS.http : tld[0].protocol;
```
mhuggins: This is referencing `PROTOCOLS.http`, but it looks like it should be `PROTOCOLS.HTTP`.

src/renderer/browser/values/protocols.js
```diff
@@ -0,0 +1,11 @@
...
+  NOS: {
+    tld: '.neo',
+    protocol: 'nos:'
+  }
+};
+
+export const PROTOCOLS = {
+  HTTPS: 'https:',
+  HTTP: 'http:'
+};
```
mhuggins: I would just export these directly since we don't ever do anything with the object directly, i.e.:

```js
export const HTTPS = 'https:';
export const HTTP = 'http:';
```

Then you can just import via:

```js
import { HTTP, CUSTOM_PROTOCOLS } from '../values/protocols';
```

.eslintrc.json
```diff
@@ -14,6 +14,7 @@
     "__static": true
   },
   "rules": {
+    "linebreak-style":"off",
```
mhuggins: The `--fix` option for eslint should allow us to not need this rule.

DalderupMaurice: woops 🙈 I always enable this because I get this issue in windows in every file/line with the difference of linestyle :P forgot to turn it off

src/renderer/browser/util/parseURL.js
```diff
@@ -1,7 +1,12 @@
...
 
 export default function parseURL(query) {
-  const trimmedQuery = trim(query);
-  const formattedQuery = trimmedQuery.includes('://') ? trimmedQuery : `nos://${trimmedQuery}`;
-  return new URL(formattedQuery);
+  const trimmedQuery = trimEnd(trim(query), '/').split('://').pop();
+
+  const tld = filter(CUSTOM_PROTOCOLS, (protocol) => endsWith(trimmedQuery, protocol.tld));
+  const tldToApply = isEmpty(tld) ? HTTP : tld[0].protocol;
+
```
mhuggins: There is a very minor issue with this where the user could enter "https://foo.neo", but we redirect to "nos://foo.neo".    It's only exacerbated because we're splitting on all `://` and not just the first.  I think we can resolve this pretty easily though.  I'll throw up an extra commit of what I think this looks like so we can get this merged.

Issue #489, opened 2018-08-29
Several issues with current toasts


DalderupMaurice:
// NOTE - removed expected/current behavior as I have multiple issues/suggestions.

- [ ] You cannot press log out button when a toast is active as the notification spreads across the whole width of the screen, over the button
- [ ] You cannot close notifications on the login screen. With this, you can't even click any modal in the login screen (import faulty wallet file still shows a modal, which you cannot close because of this issue)

## Steps to Reproduce
1. Uhh, try out as described :P 

## Possible Solution
???

## Additional context
//

## Your Environment
* nOS Client version used: Any version including toasts, I've been experiencing some of these issues a long time so they've been there a while already
* Environment name and version
    * Node version: any
    * NPM version: any
    * Yarn version: any
* Operating System and version (desktop or mobile): any

mhuggins:
For the first point, we can probably just add `user-select: none` to the parent div, and `user-select: auto` to each individual toast.

DalderupMaurice:
Doesn't work - It's related to the absolute position & having top, bottom, right, left set to 0
If I turn any of these 5 properties off, they're clickable

jeroenptrs:
@DalderupMaurice can we close this issue now?

DalderupMaurice:
These only solved a part of the issues (they're not listed here anymore)
The ones shown here are still unresolved

DalderupMaurice:
closing - first issue is going to be obsolete for 0.6, 2nd is solved

PR #490, opened 2018-08-29
chore(toasts): replaced modals with toasts, fixed network settings

DalderupMaurice:
## Description
Network settings didn't clear props and still showed a modal + toast
Also added some extra validation as you were able to add empty entries, resulting in errors
Replaced login error modal (with faulty import of wallet file) with an error toast

## Motivation and Context
#489 Although this PR doesn't cover all issues with the toasts

## How Has This Been Tested?
manual testing, shall I write tests? 

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #

### DIFF
--- a/src/renderer/settings/components/NetworkSettings/NetworkSettings.js
+++ b/src/renderer/settings/components/NetworkSettings/NetworkSettings.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { func, string, object, arrayOf } from 'prop-types';
 import { settings } from '@cityofzion/neon-js';
-import { noop, map } from 'lodash';
+import { map } from 'lodash';
 
 import LabeledInput from 'shared/components/Forms/LabeledInput';
 import LabeledSelect from 'shared/components/Forms/LabeledSelect';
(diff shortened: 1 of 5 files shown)


src/renderer/settings/components/NetworkSettings/NetworkSettings.js
```diff
@@ -121,8 +123,15 @@ export default class NetworkSettings extends React.PureComponent {
       return element.name === this.props.networkName;
     });
 
+    if (!this.props.networkName || !this.props.networkUrl) {
```
mhuggins: Should we use `isEmpty` here too?

DalderupMaurice: Cool call xD 
Btw, I think we can ditch "withAlert" hoc, right? As it's completely replaced by the toasts. If so, I could include that aswell

mhuggins: Let's keep it, because we will probably/possibly need alerts for things even if not the place we're using toasts.

Issue #133, opened 2018-05-09
Show GAS fee on real invocation confirmation dialogue


deanpress:
## Expected Behavior
Since invocations above 10 GAS aren't free, users should be made aware of the fee when sending an invocation.

## Current Behavior
The invocation dialogue is currently not displaying any fees when reviewing real invocations.

## Possible Solution
Before sending a real invocation, the client should do a testinvoke, then display the fee information and the "expected output" retrieved from the test invocation in the invocation confirmation dialogue.
In the case they fee is `<=10 GAS`, set displayed fee to "free".
In the case it's `>10` GAS, display `total GAS fee - 10` as the displayed fee.

jeroenptrs:
@deanpress one way would be to have a `testinvoke` go before an invoke, where you can see the calculated gas cost. However, we should test how this works with SC's using CheckWitness, as far as I can remember that might give incorrect results when it comes to testinvoke if addresses are badly formatted.

morhafyousef:
No issue

Issue #118, opened 2018-04-29
Build confirm function wrapper for complex dialogs


mhuggins:
## Expected Behavior
The `confirm` redux action needs a wrapper to support more complex form dialogs.  It should support complex interactions (more than just a string for its `children`) as well as being able to pass a result as arguments to `onConfirm`.

## Current Behavior
The `confirm` redux action accept children string/node and an `onConfirm` callback that is not given any arguments.

## Possible Solution
🤔

## Steps to Reproduce (for bugs)
1. Checkout #109 & run client
2. Navigate to nOS settings
3. Click "Add custom network configuration"
4. Observe the complex dialog interaction

## Context
https://github.com/nos/client/pull/109#discussion_r184884800

## Your Environment
* Version used: all
* Environment name and version (e.g. node.js 9.6.1, yarn 1.6.1, npm 5.6.0): all
* Operating System and version (desktop or mobile): all

jeroenptrs:
I think this is a bit too complex for a `good first issue` tag, which is why I'm removing it.

mhuggins:
An example of how this can be more encapsulated can be seen in #513, where it is wrapped in an `InvocationPrompt` component.  The same could be done for any modal in the future.

PR #513, opened 2018-09-07
feat(invoke): invocation modal

### LINKED ISSUE #369: Confirmation modal text does not remain inside container
## Expected Behavior
Text should be contained inside the modal confirmation window

## Current Behavior
image

## Steps to Reproduce
1. Go to https://wifeofxardas.github.io/ and try to create a new lot
2. Bid on said lot

## Your Environment
* nOS Client version used: dev latest
* Operating System and version (desktop or mobile): Windows 10

### LINKED ISSUE #474: Invocation Modal
The invocation modal should match the designs.  There should also be a success modal that displays once the invocation is processed & accepted by the blockchain.

Implementing new designs.  This is necessary to ensure the user has enough insight regarding what they are sending.

Depends on #88 & #118.

mhuggins:
## Description
This creates a new `withInvocationPrompt` HOC that is used when confirming an invocation.  The messaging depends on whether `assets` have been attached to the invocation.

## Motivation and Context
The user should have clear messaging for invocations, especially if it will cost them anything.

## How Has This Been Tested?
Smoke tested against a simple app created with create-nos-app.

screen shot 2018-09-07 at 6 29 08 pm

screen shot 2018-09-07 at 7 39 09 pm

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #369 
Fixes #474


DalderupMaurice:
Where/when would this window pop up? @mhuggins

DalderupMaurice:
- Invocations without assets: Looks good
- Invocation with assets: You say `Transfer` twice. Maybe have something like `Attach 0.3 GAS` to perform...
- Sending assets: asset ICON did not display - but that's probably out of the scope for this PR :P

src/renderer/account/components/TransactionsPanel/Send/Send.js
```diff
@@ -91,8 +91,13 @@ export default class Send extends React.PureComponent {
     const { confirm, receiver } = this.props;
     const symbol = this.getSymbol();
 
-    confirm(`Would you like to transfer ${this.getAmount()} ${symbol} to ${receiver}?`, {
-      title: 'Confirm fund transfer',
+    confirm((
+      <span>
+        Would you like to transfer {this.getAmount()} {symbol} to address{' '}
```
DalderupMaurice: does a <br /> not work?

deanpress:
Re:
>@DalderupMaurice: Where/when would this window pop up in the future? @mhuggins

I can imagine this showing up when clicking on a toast that shows when the invocation was confirmed on chain

mhuggins:
This is the modal that a user sees when the invoke API function is called.
There's a separate modal the user sees when it's on chain.  That one is in
figma, it's not in the client yet.

src/renderer/browser/hocs/withInvocationPrompt/InvocationPrompt.js
```diff
@@ -0,0 +1,110 @@
...
+
+  getSymbol = (scriptHash) => {
+    const balances = this.props.balances || {};
+    const { symbol } = balances[scriptHash] || {};
+    return symbol;
+  }
+
+  getIcon = (scriptHash) => {
+    const balances = this.props.balances || {};
+    const { icon } = balances[scriptHash] || {};
```
DalderupMaurice: you cannot destructure icon from an empty object? 
Rather initialize it like this: `const { icon = {} } = balances[scriptHash];`
some for line 101

mhuggins:
@DalderupMaurice 

> Invocation with assets: You say Transfer twice. Maybe have something like Attach 0.3 GAS to perform...

Yeah, I took that [from the designs](https://www.figma.com/file/apqKpVZZDzppAb0lVH69iiZn/nOS-Client?node-id=47%3A3390).  The modal title is "Transfer" and then we say "Transfer 0.3 GAS", is that what you mean?  I think the subtitle should be as succinct as possible, especially since multiple assets could be attached, e.g.: "Transfer 12 NEO and 0.00000005 GAS".

src/renderer/account/components/TransactionsPanel/Send/Send.js
mhuggins: I wanted to let it flow as a sentence since different screens might render differently, especially if zooming is different.

src/renderer/browser/hocs/withInvocationPrompt/InvocationPrompt.js
mhuggins: Those aren't quite the same.  In my code, `icon` ends up being `null` if the scripthash doesn't exist inside of `props.balances`, but in your code it ends up being `{}`.

mhuggins:
@DalderupMaurice I added some validation to ensure that we never open the invocation modal with invalid arguments (esp. invalid assets shape).  This should help to ensure we don't get into a state where the asset icon can't be displayed properly.

The validation accounts for invalid type (not an object), invalid asset IDs (not NEO or GAS ID), and invalid amounts (decimal for NEO values, 9+ digits for GAS).

validate-args

DalderupMaurice:
> It should have displayed if the GAS or NEO asset ID was passed in (without "0x" prefix). I'll do some follow-up work to ensure we're validating the assets data on invocations, and to reject the API request if validation fails. I think this should help to prevent an issue where a bad ID or amount is provided.
I meant, sending assets through the wallet section :P

And I think option 2 would work best

DalderupMaurice:
Little note - do take in consideration we need to show the GAS cost in the future. This would require the designs to be refactored https://github.com/nos/client/issues/133

DalderupMaurice:
I think we can enclose this issue in this PR aswell? https://github.com/nos/client/issues/118

mhuggins:
Good call on the GAS transfer fee.  Looks like we have a story to track that already, we just need to do a testinvoke first I think.

PR #535, opened 2018-09-15
feat(balance): update balances on new blocks

### LINKED ISSUE #475: Refresh account
The account assets/tokens should automatically update on new blocks.

The account state does not update.

We will similarly be doing block polling as part of #54.  This should update as part of the same poll.

## Additional context
N/A

mhuggins:
## Description
This adds automatic refreshing for account balances whenever a new block is received.

## Motivation and Context
The account screen currently does not update and has no mechanism for updating, not even when sending assets.

## How Has This Been Tested?
* Opened the account screen, and observed a single balances data fetch.
* Watched for new blocks being successfully received in redux store.
* Observe balances data re-fetch.
* Ensure that account screen doesn't change to "Loading" messaging between fetching.
* Ensure that account screen does change to "Loading" messaging when changing networks.

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #475.

### DIFF
--- a/__tests__/renderer/root/components/Layout/AuthenticatedLayout/AuthenticatedLayout.test.js
+++ b/__tests__/renderer/root/components/Layout/AuthenticatedLayout/AuthenticatedLayout.test.js
@@ -2,12 +2,16 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { MemoryRouter } from 'react-router-dom';
 
-import { provideState } from 'testHelpers';
+import { provideState, spunkyKey, mockSpunkyLoaded } from 'testHelpers';
 
 import AuthenticatedLayout from 'root/components/AuthenticatedLayout';
 import { EXTERNAL } from 'browser/values/browserValues';
 
 const initialState = {
+  [spunkyKey]: {
+    currentNetwork: mockSpunkyLoaded('MainNet'),
+    auth: mockSpunkyLoaded({ address: 'ALfnhLg7rUyL6Jr98bzzoxz5J7m64fbR4s' })
+  },
   browser: {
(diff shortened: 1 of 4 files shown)


DalderupMaurice:
I launched the client with `nOSLocal` selected from the beginning while it was disconnected, switched network, did a transaction and it's not updating because it's still polling nOSLocal.

DalderupMaurice:
Side note* 
adding new network DOES successfully update the block info thing
Changing networks does not.

mhuggins:
@DalderupMaurice I just fixed that issue here.  I rebased the branch so you'll need to re-pull / re-checkout.

mhuggins:
> What's the reasoning behind the polling time?  Could we increase this a little (15sec?), console/redux console is getting spammed and I think 5sec is a little too much if the userbase increases.

I just picked something that was smaller than the block time so that dApps could be notified about all blocks.  We can arguably raise it.  Redux console updating frequently is not a concern, but possibly DDoS'ing RPC servers is -- though we have consensus nodes to help prevent that.  Maybe we could do 10s polling for now?

> Also, it would be great if you could move the polling time to the values folder. This was missed in the previous PR, merged a little too early, Jeroen missed that.

I try to only put value constants into other files if they're shared across files, otherwise it seems like an extra hoop to jump through to find a value.  What do you think about that?

mhuggins:
I updated to 10s for now, but we can change at any time given our needs.

mhuggins:
> Amazing work 😄 (note the height in each console, switched networks in between)

This makes me think we should add a "network" event too for when the user changes network, as well as a `getNetwork` API function to get the currently selected network.

PR #538, opened 2018-09-15
feat(browser): add app menu and browser navigation controls

### LINKED ISSUE #231: Add back, forward and reload buttons to browser
## Current Behavior
None of these controls are available. It makes it impossible to navigate through history.

## Possible Solution
Extend the browser bar and add new components. Actions should control the webview component:
https://electronjs.org/docs/api/webview-tag#methods (e.g. `<webview>.canGoBack()`, etc.)

mhuggins:
## Description
This adds:

* Application menu options for the browser:
  * View -> Stop
  * View -> Reload
  * View -> Toggle Developer Tools
  * History -> Back
  * History -> Forward
* Address bar navigation functionality:
  * Back
  * Forward
  * Reload

## Motivation and Context
We want the basic UX to be implemented before our 0.3.0 release.  This gets us much closer.

## How Has This Been Tested?
Smoke tested opening multiple tabs, opening account settings, checking to ensure the menu state updates properly when changing tabs, when navigating back/forward, that performing an action only pertains to the active tab, etc.

I made a gif, but it's too big. 😞

screen shot 2018-09-15 at 3 32 12 pm

screen shot 2018-09-15 at 3 32 16 pm

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #231.


src/renderer/browser/components/DAppContainer/DAppContainer.js
```diff
@@ -23,17 +25,18 @@ export default class DAppContainer extends React.PureComponent {
...
 
   static defaultProps = {
-    className: null
+    className: null,
+    active: false,
+    onFocus: noop
   }
 
   async componentDidMount() {
-    window.addEventListener('focus', this.handleFocus);
```
mhuggins: Whoops, I didn't mean to remove this.  Needs to be added back in before merging.

DalderupMaurice:
Can you disable the forward/back button when you cannot go forward/backwards, the same as you did with the context menu

Also, windows doesn't have the Application menu window :P

DalderupMaurice:
Another note, is it possible to have the context menu on the error page?

mhuggins:
> Can you disable the forward/back button when you cannot go forward/ backward, the same as you did with the context menu

I'd like to punt on this and do it in a follow-up PR.  I need to figure out a good way forward on this, because binding webview events to components is not super easy (requires IPC communication to delegate webview events).  I'd rather get this in for now since it's a major improvement for 0.3.0, and follow up on the second aspect of it for a minor improvement.

> Also, windows doesn't have the Application menu window :P So I cannot test that

Shit, I'll check it out.  Thanks!

> One more thing, I think the default for navigating for windows/linux is ALT + arrow key instead of CTRL + Arrow key

This should be an easy fix, I can update.

DalderupMaurice:
> Also, windows doesn't have the Application menu window :P So I cannot test that
I think this is better for a follow-up PR as well, as it's quite "huge". @mhuggins  
It would require a dedicated menu bar (or another option?) which would only be available for windows. 
As of now, we do not have designs for that yet, and I am not sure if it would look pretty (chrome etc doesn't have a bar either on windows, but it does have one for MacOS)

mhuggins:
> Sets menu as the application menu on macOS. On Windows and Linux, the menu will be set as each window's top menu.

Ahh yep, that's from the [Menu.setApplicationMenu electron documentation](https://electronjs.org/docs/api/menu#menusetapplicationmenumenu).  Since we're hiding the menus on windows directly, makes sense that we shouldn't add this now.

DalderupMaurice:
Test is failing aswell :P

mhuggins:
@DalderupMaurice Just pushed an update to fix tests and change Windows back/forward to use Alt+ArrowKey.

src/renderer/root/components/AuthenticatedLayout/AddressBar/index.js
```diff
@@ -25,6 +26,9 @@ export default compose(
   withHandlers({
     onQuery: (props) => (target) => {
       props.navigate(props.activeSessionId, target);
-    }
+    },
+    onBack: (_props) => () => ipcRenderer.send('history:back'),
```
DalderupMaurice: aren't these `_props` unused? Or am I missing a piece :P

PR #572, opened 2018-09-25
Feat/priority fee

### LINKED ISSUE #392: Add ability to set GAS priority fee for transactions
Users should be able to set a GAS priority fee for making NEO-based transactions. There should be a configurable default in Settings, and it should be configurable in the transaction/invocation modal.

When the free tx pool is full, additional free transactions get declined. Adding a small tx fee will solve this, so we should give users the ability to configure this.

mhuggins:
**NOTE**: Marking as `don't merge` since the fee doesn't seem to get applied even though it's being passed to neon-js properly.  Waiting for some feedback from snowypowers before this needs a review.

## Description
This adds a priority fee to transactions (contracts and invocations).

## Motivation and Context
Network congestion will cause transactions without a priority fee attached to not get sent quickly.

## How Has This Been Tested?
Sending GAS via the account screen.

priority-fee

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #392.


src/renderer/browser/components/RequestsProcessor/Send/index.js
```diff
@@ -48,19 +50,21 @@ export default function makeSend(sendActions) {
...
+    withInitialCall(sendActions, ({ net, amount, asset, receiver, address, wif, remark, fee }) => ({
       net,
       amount,
       asset,
       receiver,
       address,
       wif,
-      remark
+      remark,
+      fee
```
DalderupMaurice: You seem to fix the usage of fee/fees throughout the whole codebase. Could we use 1 constant for this? :P

mhuggins: There's some cut-off point where we refer to it as "fee" but track it as "fees".  For example, in the UI we call it "Priority Fee".  That was the what I used as the decision for us to store it as `fee` instead of `fees`.  It's only when it gets into neon-js that it gets converted to `fees`.

I would prefer we use language that represents nOS usage vs. a what a third party library prefers, especially considering the library can change at any given time.

mhuggins:
@DalderupMaurice Ahh yeah, I actually fixed this because it's the intended layout, but it broke somewhere along the way.  I'll see if we can size it better for larger screens.

mhuggins:
@DalderupMaurice Updated to fix the merge conflict.  I'm going to merge to unblock the next release, but the settings form sizing should be an easy follow-up fix.

PR #585, opened 2018-09-27
add color to the comments

Nikolaj-K:
prettification

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,11 +37,12 @@
 
 # Getting Started with nOS
 
-The standalone client installers can be found [here](https://github.com/nos/client/releases).
+The standalone client installers can be found 
+[here](https://github.com/nos/client/releases).
 To build manually, the client can be cloned from GitHub and run using the development steps below.
 
 ### Commands
-```
+```c
 // Cloning from Github
 git clone https://github.com/nos/client.git
 
@@ -62,7 +63,9 @@ yarn dist
 
 
 ## Getting started with nOS Development
-Need some help with building nOS dApps? Check out our dedicated [documentation website](https://docs.nos.io/) containing useful info for the following topics:
+Need some help with building nOS dApps? Check out our dedicated 
+[documentation website](https://docs.nos.io/) 
+containing useful info for the following topics:
 - [**nOS Client API Documentation**](https://docs.nos.io/docs/nos-client/api.html)
 - [Create nOS dApp Usage](https://docs.nos.io/docs/create-nos-dapp/installation-usage.html)
 - [nOS Local Setup and usage](https://docs.nos.io/docs/nos-local/installation-usage.html)
@@ -76,7 +79,9 @@ Future resources:
 
 
 # Contribute to this repository
-We welcome contributions to the code base. If you are interested in becoming a contributor, please read the [contributing guide](/.github/CONTRIBUTING.md) that covers the following:
+We welcome contributions to the code base. If you are interested in becoming a contributor, please read the 
+[contributing guide](/.github/CONTRIBUTING.md) 
+that covers the following:
 
 - [Reporting bugs](/.github/CONTRIBUTING.md#reporting-bugs)
 - [Suggesting enhancements](/.github/CONTRIBUTING.md#Suggesting-Enhancements)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,11 +37,12 @@
...
 # Getting Started with nOS
 
-The standalone client installers can be found [here](https://github.com/nos/client/releases).
+The standalone client installers can be found 
+[here](https://github.com/nos/client/releases).
 To build manually, the client can be cloned from GitHub and run using the development steps below.
 
 ### Commands
-```
+```c
```
mhuggins: This isn't C, so maybe `bash` or `shell` would be more appropriate?

README.md
```diff
@@ -100,7 +107,9 @@ Resources:
 
 # Releasing
 ## Windows & Linux
-We use [CircleCI](https://circleci.com/gh/nos/client) to automatically create builds based upon git tags.
+We use
```
mhuggins: The line breaks throughout here are kind of random.  It doesn't read like a sentence anymore.  Could we just include the color change itself?

README.md
Nikolaj-K: Indeed, this isn't C, but git markup bash is shit and doesn't show anything. C (and Cpp and derived languages, e.g. Java and Scala) will do the `//` correctly, but bash never helps for the build instructions. Actually, I usually go for haskell when I don't know, because that language always has nice highlighting. I think it's because of the pattern matching feature that are often tied to functional languages. 
Well, choose either aestetics: Using the right language flag or using the right color highlighting in practice. Sadly it doesn't seem to work to have both. 
Not sure if shell exist or if it does what you want - let's try. I don't think it will.

Nikolaj-K: Just checked: Both bash and shell will give you the opposite of what you want, namely highlighting a bunch of words in the comments.

README.md
Nikolaj-K: You're mistaken.

README.md
jeroenptrs: This is not the proper way to do it.

Best way is to indeed use `bash` or `sh` as highlighting target and use # as a comment identifier, so please change it to that.

Example:
```sh
# Cloning from Github
git clone https://github.com/nos/client.git

# Install dependencies and launch the nOS client
# This is used to start developing on the nOS client
yarn install && yarn start

# Testing command
yarn test

# Testing with debug (repl) command
yarn test:debug

# Distribution command
yarn dist
```

README.md
jeroenptrs: `The markup will ignore the line break in the plaintext, where it helps singling out and working on the link. This holds for many other text formatters as well btw.`

I'm not following, where does this provide any difference for github *or* text formatters in comparison to keeping it all on the same line?

Nikolaj-K: On the user side, it will look the same. On the editor side, the first looks like

---

"Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book, see **| lipsum reference link name |(https://www.lipsum.com/).** It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."

---

and the second will look like

---

"Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book, see 
**| lipsum reference link name |(https://www.lipsum.com/)**.
It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."

jeroenptrs: I don't see a problem with the first, honestly.

Nikolaj-K: I'd argue it's more work to edit if you have to search for the links, e.g. when some inevitably break. But it's just a style choice suggestion and so let's stick to the unedited way if it works for you.

PR #699, opened 2018-10-30
feat(browser): multi testinvoke

jeroenptrs:
## Description
Allows you to use a raw script or array of {scriptHash, operation, args} for testinvokes

## Motivation and Context
Chaining testinvokes (balanceOf calls, multiple non-intensive computations, ...)

## How Has This Been Tested?
Tested with a raw script and an object array of params and it all works!

## Types of changes

## Checklist

#### Documentation
- [x] Docs need to be added/updated (for bug fixes/features)

## Closing issues
Fixes #

### DIFF
--- a/src/renderer/browser/actions/makeTestInvokeActions.js
+++ b/src/renderer/browser/actions/makeTestInvokeActions.js
@@ -26,7 +26,7 @@ const testInvoke = async ({ net, scriptHash, operation, args, script, encodeArgs
     if (typeof script === 'string') {
       invokeScript = script;
     } else if (isArray(script)) {
-      script.forEach((s) => validateScriptArgs(s));
+      script.forEach(validateScriptArgs);
       invokeScript = createArrayScript(script, encodeArgs);
     }
   } else {
--- a/src/renderer/browser/actions/makeTestInvokeActions.js
+++ b/src/renderer/browser/actions/makeTestInvokeActions.js
@@ -19,15 +19,15 @@ const testInvoke = async ({ net, scriptHash, operation, args, script, encodeArgs
     validateScriptArgs({ scriptHash, operation, args });
     invokeScript = createScript(scriptHash, operation, args, encodeArgs);
   } else if (script) {
-    if (typeof script !== 'string' || !isArray(script)) {
+    if (typeof script !== 'string' && !isArray(script)) {
       throw new Error(`Invalid script: "${script}"`);
     }
 
-    if (isArray(script)) {
+    if (typeof script === 'string') {
+      invokeScript = script;
+    } else if (isArray(script)) {
       script.forEach((s) => validateScriptArgs(s));
       invokeScript = createArrayScript(script, encodeArgs);
-    } else {
-      invokeScript = script;
     }
   } else {
     throw new Error('Invalid config!');
--- a/src/renderer/browser/actions/makeTestInvokeActions.js
+++ b/src/renderer/browser/actions/makeTestInvokeActions.js
@@ -1,5 +1,5 @@
 import { createActions } from 'spunky';
-import { rpc, wallet } from '@cityofzion/neon-js';
+import { rpc } from '@cityofzion/neon-js';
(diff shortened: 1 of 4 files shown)


src/renderer/browser/actions/makeTestInvokeActions.js
```diff
@@ -1,39 +1,51 @@
...
-  if (!isArray(args)) {
-    throw new Error(`Invalid arguments: "${args}"`);
+const testInvoke = async ({ net, scriptHash, operation, args, script, encodeArgs }) => {
+  let invokeScript;
+
+  if (scriptHash && operation && args) {
+    validateScriptArgs({ scriptHash, operation, args });
+    invokeScript = createScript(scriptHash, operation, args, encodeArgs);
+  } else if (script) {
+    if (typeof script !== 'string' || !isArray(script)) {
```
mhuggins: Pretty sure you want `&&` here, not `||`.

src/renderer/browser/util/createArrayScript.js
```diff
@@ -0,0 +1,14 @@
...
+
+import encode from 'shared/util/encodeArgs';
+
+const createArrayScript = (scripts, encodeArgs) => {
+  const args = scripts.map((script) => {
+    return encodeArgs ? encode(script) : script;
+  });
+
+  // Create script
+  return Neon.create.script(args);
```
mhuggins: I don't think this is right.  I think we want something more like this:

```ruby
import createScript from 'shared/util/createScript';

const createArrayScript = (scripts, encodeArgs) => {
  return scripts.map(createScript);
};
```

That's not quite right either since we need to concatenate the scripts probably (using `sc` import from neon-js?), but the current logic looks like it's disregarding the `scriptHash` and `operation`.

src/renderer/browser/actions/makeTestInvokeActions.js
```diff
@@ -1,39 +1,51 @@
...
+      throw new Error(`Invalid script: "${script}"`);
+    }
+
+    if (isArray(script)) {
+      script.forEach((s) => validateScriptArgs(s));
```
mhuggins: :neckbeard: You could also just call `script.forEach(validateScriptArgs)`

src/renderer/browser/actions/makeTestInvokeActions.js
jeroenptrs: Fixed this!

src/renderer/browser/util/createArrayScript.js
jeroenptrs: No not really, the Neon.create.script needs to be called with an array (it takes both an object or an array of objects), I wrongfully used the entire array instead of just the args of each element.

jeroenptrs: Should be fixed with my new push @mhuggins

src/renderer/browser/util/createArrayScript.js
```diff
@@ -3,12 +3,12 @@ import Neon from '@cityofzion/neon-js';
 import encode from 'shared/util/encodeArgs';
 
 const createArrayScript = (scripts, encodeArgs) => {
-  const args = scripts.map((script) => {
-    return encodeArgs ? encode(script) : script;
+  const invokeScripts = scripts.map(({ args, ...restProps }) => {
+    return { ...restProps, args: encodeArgs ? encode(args) : args };
```
mhuggins: Yes, this is what I was getting at.  Nice fix!

mhuggins:
I like the direction of this change.  I think long term we can just change this to ONLY accept a `scripts` key for uniformity.

jeroenptrs:
@mhuggins i was thinking the same, we can slowly sunset the base scripthash/operation/args and use an array or raw string.

jeroenptrs:
@mhuggins @DalderupMaurice @deanpress everything is passing on my end, all testinvokes seem to work (base arguments, raw script and array of arguments). Can we get this baby in? 👶

PR #742, opened 2018-12-05
feat(account): send panel token/currency conversion

### LINKED ISSUE #411: Fiat to token conversion
The "Send" tab on the account screen should include a fiat input along with the token input, which can be used as an alternative mechanism for entering the amount of a token to send.  When the token or fiat value is updated by the user, the other should be calculated and updated automatically.

Simplifies sending for users with consideration for the currency they're most familiar with.

There is a user setting for preferred currency (e.g.: USD).  That currency code/symbol/format will need to be used for calculating and rendering.  There is also an existing `formatCurrency` function for rendering currency values with the appropriate commas, decimals, and symbol.

One consideration is that we will need to use rounding or a ceiling method when converting.  The "decimals" value for a token will need to be considered.  For values like NEO that have 0 decimal places, we will need to decide if entering $5 USD when the NEO value is $10 USD should map to 0 NEO or 1 NEO.

mhuggins:
## Description
This adds the token value/fiat conversion when sending assets/tokens from the send panel.

## Motivation and Context
To support currencies that the user is familiar with.

## How Has This Been Tested?
Changing the token value, changing the currency value, changing the currency type, focusing & blurring after changing the amount values.

conversion-input

## Types of changes

## Checklist

#### Documentation

## Closing issues
Fixes #411


DalderupMaurice:
Should/Could we keep the number as a whole, when it's for example: 4.00, to not add the 2 trailing zeros? 
I'll have a proper review and test it when I get home

jeroenptrs:
@DalderupMaurice For fiat inputs we should definitely keep the trailing zeroes.

DalderupMaurice:
> @DalderupMaurice For fiat inputs we should definitely keep the trailing zeroes.

It looks weird when you have the value `4` and suddenly there are 2 zeros appended

DalderupMaurice:
Another tiny thing is the following: 


NEO is suddenly divisible :P

mhuggins:
> if I insert 1 or 2 euro, it displays 0 NEO, but when I insert 3 euro, it displays 1 NEO. Is this intended?

Yes, I confirmed with Dean before starting on this that we should round down to the nearest whole number.  You can't get 1 NEO for 2 Euro, but you can for 3 Euro.

mhuggins:
> NEO is suddenly divisible

I'll get a fix for this up.

mhuggins:
@DalderupMaurice I just pushed up a fix for the issue which switching from GAS to NEO.

DalderupMaurice:
I just noticed that the imprecision is because the balance is truncated with 2 decimals and rounded up, while the calculation is rounded down. This is no problem for me, wbu? @deanpress

deanpress:
@DalderupMaurice Fine with me. Perhaps we should add a small disclaimer that fiat values are estimates?

deanpress:
@DalderupMaurice "Estimate" would also point out the fact that the exchange rate could have changed during the transaction.

PR #1118, opened 2019-08-20
feat(auth): add the ability to import existing seed

### LINKED ISSUE #1026: Add ability to restore your account from a given mnemonic seed
Have the ability for the user so restore/import their account using an existing mnemonic seed

Users with an existing mnemonic need to create a new one

/

## Additional context
/

jeroenptrs:
## Description
Import existing BIP39 seed

WIP because still need to add copy

## Motivation and Context
Must have functionality

## How Has This Been Tested?
Imported my keys

## Types of changes

## Checklist

#### Documentation
- [x] Docs need to be added/updated (for bug fixes/features)
(Potentially a tut on how to do this? 🤷‍♂ )

## Closing issues
Fixes #1026


src/renderer/auth/shapes/accountShape.js
```diff
@@ -6,6 +6,7 @@ export default shape({
   accountLabel: string.isRequired,
   secretWord: string.isRequired,
   isHardware: bool.isRequired,
+  isImport: bool.isRequired,
```
DalderupMaurice: This can be removed as it's only used/needed in the registerShape

src/renderer/register/actions/registerFormActions.js
```diff
@@ -39,14 +39,24 @@ export const validateAndStoreFormData = async (data) => {
...
 
-  // Generate bip39 Mnemonic - 256-bits entropy (24-word long mnemonic)
-  const mnemonic = bip39.generateMnemonic(256, null, bip39.wordlists[DEFAULT_LANGUAGE]);
-  const encryptedMnemonic = simpleEncrypt(mnemonic, passphrase);
-  return {
-    ...data,
-    mnemonic,
-    encryptedMnemonic
-  };
+  if (!isHardware && !isImport) {
```
DalderupMaurice: You can safely adjust this if/else if/else as mnemonic is only NOT required on the first step when isImport is true 

Instead you could do something like this: 
```js
if(isImport && !!data.mnemonic) return data;

const mnemonic = isImport 
                                    ? data.mnemonic  
                                    : bip39.generateMnemonic(256, null, bip39.wordlists[DEFAULT_LANGUAGE]);

const encryptedMnemonic = simpleEncrypt(mnemonic, passphrase);
return {
     ...data,
    mnemonic.
    encryptedMnemonic
};
```
(not 100% if that's correct :P)

src/renderer/register/components/AccountView/ImportView/ImportView.js
```diff
@@ -0,0 +1,111 @@
...
+    mnemonic: arrayOf(string).isRequired,
+    setMnemonic: func.isRequired
+  };
+
+  static defaultProps = {
+    loading: false
+  };
+
+  state = {
+    validMnemonic: false
```
DalderupMaurice: I personally always use `withState` from recompose in the index file instead of local state. (refer to setMnemonic)

src/renderer/register/components/AccountView/ImportView/ImportView.js
```diff
@@ -0,0 +1,111 @@
...
+  };
+
+  mnemonicRefs = Array(24)
+    .fill(undefined)
+    .map(() => React.createRef());
+
+  showErrorToast = debounce((e) => {
```
DalderupMaurice: Why is this debounce required? I'm not sure what the resulting behavior will be

src/renderer/register/components/AccountView/ImportView/ImportView.js
```diff
@@ -0,0 +1,111 @@
...
+    const { mnemonic: mnemonicArray, account, storeFormData } = this.props;
+    const mnemonic = mnemonicArray.join(' ');
+
+    storeFormData({
+      ...account,
+      mnemonic
+    });
+  };
+
+  setMnemonic = (count) => async (event) => {
```
DalderupMaurice: doesn't have to be async I guess?

src/renderer/register/components/AccountView/ImportView/MnemonicWord/index.js
```diff
@@ -0,0 +1,50 @@
+import classNames from 'classnames';
```
DalderupMaurice: can't we reuse this as well for the mnemonic overview, when import is not selected? This also has a class "MnemonicWord" if I remember correctly

src/renderer/register/components/AccountView/ImportView/ImportView.js
jeroenptrs: So it doesn't pop up on every key press

src/renderer/register/components/AccountView/ImportView/ImportView.js
jeroenptrs: So it doesn't pop up on every key press

src/renderer/register/actions/registerFormActions.js
jeroenptrs: This wouldn't work since data gets returned before encrypting the mnemonic

jeroenptrs: I see your point though. Will try out something

src/renderer/register/components/AccountView/ImportView/ImportView.js
jeroenptrs: Yes but recompose state handlers don't accept callbacks. Callback functionality

src/renderer/register/components/AccountView/ImportView/MnemonicWord/index.js
jeroenptrs: I don't see why 🙈 this is styling for an input form

src/renderer/register/components/CreateAccount/RegisterForm/RegisterForm.js
```diff
@@ -96,34 +101,62 @@ export default class RegisterForm extends React.PureComponent {
...
       </form>
     );
   };
 
-  renderCheckboxLabel = () => {
-    return <Label htmlFor="isHardware" label="Use Ledger for next step" className={styles.label} />;
+  renderButtonMessage = () => {
+    const { isHardware, isImport } = this.props;
+    if (isHardware) return 'Next: connect Ledger';
+    if (isImport) return 'Next: import seed';
```
deanpress: Button labels should have each first word capitalized

